PR #34, opened 2022-07-18
rm decache

psugihara:
Seems to work much better without this because it doesn't unrequire react. My bad!


packages/cli/src/preview/controllers/previews.ts
```diff
@@ -1,12 +1,11 @@
+import React from "react";
```
alexfarrill, line 1: necessary?

psugihara: don't think so! rm'd

alexfarrill (approved):

PR #36, opened 2022-07-18
Gh/27 show mjml errors

alexfarrill:
@monicakogler will you give some feedback on making the error state look better.  Thanks!

image


alexfarrill:
Unfortunately the line numbers indicated only directionally correspond to the true line number

psugihara:
Hm weird, maybe it's the line number from the compiled mjml rather than the react-mjml.

PR #43, opened 2022-07-20
fetchPreview rather than hard reload; fixes #42

### LINKED ISSUE #42: remove hard reload
Spotted a small regression with the change to make error messages display. It's doing a hard reload rather than refetching the html. This makes it feel a bit slower because it has to do 2 fetches and more rendering. Instead, let's re-trigger the fetch and just reload the iframe content.

psugihara:


psugihara:
merging to put in the vid :)

PR #47, opened 2022-07-22
Add cypress and some basic e2e tests

alexfarrill:
🤗


psugihara (approved):
nice, we can follow up with adding this to CI

PR #53, opened 2022-07-26
fixes for react 17 type errors

psugihara:
This was breaking when I made a remix app with `npx create-remix@latest`. Maybe the remix app I had tested was react 18?

It would be cool to have cypress smoke tests for multiple react versions.

### DIFF
--- a/packages/cli/src/generator_templates/ts/emails/components/Head.tsx
+++ b/packages/cli/src/generator_templates/ts/emails/components/Head.tsx
@@ -13,11 +13,12 @@ type HeadProps = { children?: ReactElement };
 const Head: React.FC<HeadProps> = ({ children }) => {
   return (
     <MjmlHead>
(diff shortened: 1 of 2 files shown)


packages/cli/src/generator_templates/ts/emails/components/Head.tsx
```diff
@@ -13,11 +13,12 @@ type HeadProps = { children?: ReactElement };
 const Head: React.FC<HeadProps> = ({ children }) => {
   return (
     <MjmlHead>
-      <MjmlFont
-        name="Inter"
-        href="https://fonts.googleapis.com/css2?family=Inter:wght@400;600;900"
-      />
-      <MjmlStyle>{`
+      <>
```
psugihara, line 16: expects a single child

alexfarrill (approved):

PR #57, opened 2022-07-26
JS support

### LINKED ISSUE #37: js support
At the moment mailing only integrates well in a ts project. I removed the JS support in #30 because it wasn't working. Requiring jsx templates was blowing up on es6 imports.

Plan:
- bring back js generators and init prompt from #30 
- figure out how to make es6 `import` working

psugihara:
fixes #37 🎊 🥳 

redwood js and ts ✅ 
next js and ts ✅ 
remix js and ts ✅


.changeset/pre.json
```diff
@@ -0,0 +1,17 @@
+{
+  "mode": "pre",
+  "tag": "next",
```
psugihara, line 3: going to try exiting pre on this branch and fixing up changesets

🧠 https://github.com/changesets/changesets/blob/main/docs/prereleases.md

psugihara: kk seems good now, bumped all to 0.5.0!

PR #66, opened 2022-08-01
include mdx and demo blog post

alexfarrill:
- Include mdx library
- demo of blog post using mdx at http://localhost:3000/blog

image

@psugihara this is rendering but all styles are reset probably by one of the style libraries you included.  Can you show me what is doing that or what other things should be included on the page?

Todo:
- @psugihara for future, do we want dynamic routing for blog posts instead of adding each to pages-manifest.json?


psugihara (commented):
Nice!

Styles are added in a few ways:
- global css [imported in _app](https://github.com/successor-software/mailing/blob/main/packages/web/pages/_app.js#L1)
- the css module like you're using 
- tailwind, which was added like this: https://tailwindcss.com/docs/guides/nextjs

I pulled your branch and it looks like those do all work in the mdx file. Tailwind has the css reset.

Maybe it would be good to use MDXProvider to style the markdown components like they do here:

Screen Shot 2022-08-01 at 3 15 57 PM

packages/web/pages/blog/index.mdx
```diff
@@ -0,0 +1,11 @@
+import styles from "../../styles/Home.module.css";
+
+<>
+
+<main className={styles.main}>
```
psugihara, line 5: This style doesn't exist, so it just gets stripped out of the html. Probably best to just remove.

psugihara:
> @psugihara for future, do we want dynamic routing for blog posts instead of adding each to pages-manifest.json?

I don't think we need dynamic routing... we can just drop the posts in that directory with the slug as filename.

alexfarrill:
Screen Shot 2022-08-02 at 3 52 50 PM

@psugihara Why are the netlify builds failing?

alexfarrill:
nm I think I have to move the components folder up

psugihara (approved):
sick, looks great!

agamm:
@psugihara So can I use MDX by default now?

psugihara:
@agamm no sorry, this was for the website [mailing.run](https://www.mailing.run)

it's a cool idea though. If that's something you want, feel free to open an issue or discussion with some more detail on your use-case and how you might want to use and customize it.

agamm:
@psugihara done https://github.com/sofn-xyz/mailing/issues/150

Issue #31, opened 2022-07-18
easily deploy previews to web for design QA/demo purposes (e.g. `mailing build` -> static site)


psugihara:
command could also be `mailing export` (might be better since this is not required for production)

psugihara:
https://mailing.run

PR #68, opened 2022-08-02
freeze script for building static version of UI

psugihara:
This adds a cleaned up version of the functionality to generate https://demo.mailing.run
Just want to get it into main so it stays up to date.

- ./scripts/freeze passes an environment variable NEXT_PUBLIC_STATIC so that the app knows it's 

Closing #31 in favor of this for now. In the future it would be cool to expose this functionality to users.


psugihara:
hm, seeing a problem with demo.mailing.run that I don't see locally:

- go to demo.mailing.run
- click into preview
- hit back in browser
- expected: back at index
- actual: white screen

psugihara:
kk works

alexfarrill (approved):

PR #77, opened 2022-08-03
No react imports

### LINKED ISSUE #73: Running without importing React?
Hey! Loving this project, this is exactly what I was after for a long time (https://github.com/wix-incubator/mjml-react/issues/37). I am playing around with the boilerplate and I discovered a few issues I will report in this project. The first one is whether it's possible to run this project without importing React? I copy-pasted my tsconfig from another project that supports this, but running `mailing preview` after removing all React imports throws this error

```
error TS2686: 'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.
```

psugihara:
This fixes #73 by switching to the modern automatic jsx runtime.

Redwood, next, and remix all support this.

tested ts and js

Also, cleaned up some over-logging that babel and next were causing:
Screen Shot 2022-08-03 at 1 25 21 PM
Screen Shot 2022-08-03 at 1 45 23 PM

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,6 +1,6 @@
 {
   "presets": [
-    "@babel/react",
+    ["@babel/react", { "runtime": "automatic" }],
     ["@babel/preset-typescript", { "isTSX": true, "allExtensions": true }],
     "@babel/preset-env"
   ]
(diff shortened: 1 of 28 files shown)


alexfarrill (approved):

mrlubos (commented):
Thanks for the quick turnaround on this one!

PR #84, opened 2022-08-04
dev symlinks templates so you don't have to copy/past changes to gene…

psugihara:
…rators

This improves the dev scripts. It symlinks the templates into the project root rather than copying so that when you make changes, they are saved in the generators without any change to your workflow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


monicakogler:
love it. is this only possible with ts?

psugihara:
no it works with both (`yarn devjs` to link the js templates in emails)

monicakogler (approved):

PR #85, opened 2022-08-04
Dark mode emails

monicakogler:
updates js and ts components and templates to support dark mode


packages/cli/src/preview/controllers/previews.ts
```diff
@@ -56,8 +56,9 @@ export function showPreview(req: IncomingMessage, res: ServerResponse) {
   const functionName = functionNameJSON.replace(".json", "");
 
   delete require.cache[modulePath];
-  const module = require(modulePath);
-  const component = module[functionName]();
+  console.log(modulePath, require.cache)
+  const templateModule = require(modulePath);
+  const component = templateModule[functionName]();
```
psugihara, line 61: remove this change plz (this was the debugging i did on your comp last night)

PR #95, opened 2022-08-06
Html UI

monicakogler:
header refresh

Screen Shot 2022-08-05 at 8 03 40 PM

also:
- fixes nextjs eslint warnings
- adds HTML viewing to preview UI


psugihara (approved):
think this is good but had to make some more changes to get images to work in the static export (see demo links from vercel)

psugihara:
tagging @alexfarrill for another look

psugihara:
got impatient / wanted in my branch :) can fix more things post merge if u see anything

PR #94, opened 2022-08-05
smoke tests for next.js

alexfarrill:
fixes #4

PR #97, opened 2022-08-07
remove js build files from git

psugihara:
We don't use these in development anymore and they kludge up the diffs.

https://github.com/sofn-xyz/mailing/pull/94#discussion_r939461053


monicakogler (approved):
cool

PR #110, opened 2022-08-09
Smoke test: add remix tests and cache

### LINKED ISSUE #107: smoke tests: add remix support

### LINKED ISSUE #108: smoke tests: use cached installs of create-framework apps for speed
this will make the tests a lot faster: 
- [ ] store a fresh install of each framework in a runs/cached directory
- [ ] write the md5 to a value in the framework app class
- [ ] compare to an MD5 and copy the fresh install if it matches
- [ ] if it doesn't match, do a fresh install and write it to the cache
- [ ] add an override option to bust and refresh the cache when we want to fetch an updated copy of the framework

alexfarrill:
fix #107
fix #108

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,7 @@ emails
 !packages/cli/src/generator_templates/js/emails/
 
 scripts/e2e_test/runs/*
+scripts/e2e_test/cache/*
 
 # Mac
 *.DS_Store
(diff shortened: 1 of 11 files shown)


psugihara (approved):
awesome, looks good! so no actual problem with remix integration?

alexfarrill:
That's right.  There was a bug in here that will let the mailing process keep running and my mailing that was running had an old version of TextEmail

PR #116, opened 2022-08-12
Gh/74 mailing config

alexfarrill:


packages/cli/src/config.ts
```diff
@@ -0,0 +1,35 @@
...
+  return !!(pkg.devDependencies?.typescript || pkg.dependencies?.typescript);
+}
+
+export function writeDefaultConfigFile(): void {
+  const fs = require("fs");
+  const prettier = require("prettier");
+
+  if (!existsSync(MAILING_CONFIG_FILE)) {
+    const configJsonString = prettier.format(
+      JSON.stringify({
```
alexfarrill, line 22: @psugihara it currently is setup to write a default config file with 2 options specified: "typescript" and "emailsDir".  Are there other options that should be set in the default config file that it generates?  e.g. 'quiet' and 'port' could be added.  should it include all options?  in any case i'll add documentation that specifies all the options

psugihara: nice, this looks good. 

`outDir` option would be good to add (used in export-previews)

packages/cli/src/config.ts
```diff
@@ -0,0 +1,35 @@
...
+        typescript: looksLikeTypescriptProject(),
+        emailsDir: "./emails",
+      }),
+      { parser: "json", printWidth: 0 }
+    );
+
+    try {
+      writeFileSync(MAILING_CONFIG_FILE, configJsonString);
+    } catch (err) {
+      console.error(err);
```
psugihara, line 32: switch to [error fuction](https://github.com/sofn-xyz/mailing/blob/main/packages/cli/src/log.ts#L9)

packages/cli/src/config.ts
psugihara: I think we should remove port as an option for now actually. It's hard coded in a few places, so bugged that we allow it. We could fix and re-add later but I don't think it really matters to have it be configurable rn.

mrlubos: Can we support `PORT` environment variable?

packages/cli/src/index.ts
```diff
@@ -1,9 +1,24 @@
+import { existsSync } from "fs-extra";
 import "dotenv/config";
 import yargs from "yargs/yargs";
 import * as preview from "./commands/preview";
 import * as init from "./commands/init";
+import { looksLikeTypescriptProject, MAILING_CONFIG_FILE } from "./config";
 
+const fs = require("fs");
```
psugihara, line 8: switch to import syntax plz

packages/cli/src/index.ts
```diff
@@ -1,9 +1,24 @@
...
+
+const config = existsSync(MAILING_CONFIG_FILE)
+  ? JSON.parse(fs.readFileSync(MAILING_CONFIG_FILE))
+  : {};
+
+// prettier-ignore
 yargs(process.argv.slice(2))
+  .boolean(['quiet', 'typescript'])
+  .config(config)
+  .default('port', 3883)
```
mrlubos, line 18: Can this use the `DEFAULT_PORT` constant from `packages/cli/src/commands/preview.ts`?

packages/cli/src/config.ts
psugihara: yeah, it's just some work to get it going. In particular, I think next will error if a path like this is not hard-coded, so we'd need to figure out a way around that: https://github.com/sofn-xyz/mailing/blob/main/packages/cli/src/pages/previews/%5BpreviewClass%5D/%5BpreviewFunction%5D.tsx#L16

packages/cli/src/index.ts
alexfarrill: Weirdly this function behaves differently when imported (returns a stream instead of a string). I will muck with it a bit more to see what is the matter with it

packages/cli/src/index.ts
alexfarrill: This should actually supercede DEFAULT_PORT, but still allow you to pass a --port option to preview and pick it up correctly

alexfarrill:
✅ e2e tests pass
✅ jest tests pass

Things to check:
- mailing init generates a mailing.config.json file for you
- upgrading in an existing project does not break anything and inherits your choices for emailsDir and typescript
- --help works for every command and shows useful info, including
- options described by --help do what they say
- cli values take precedence over config values take precedence over program defaults (see DEFAULTS in config.js)

N.B. --port option is still WIP, will continue work on it in #117

psugihara (approved):
working great now. huge upgrade.

mrlubos:
This works beautifully. Any thoughts on renaming `emailsDir` to `sourceDir` or similar? In CLI it could then be `--src`, `--files` or something along those lines. I just find the `emailsDir` naming very non-standard and tied to the package functionality, which is emails.

psugihara:
`src` is nice!

`emailsDir` is so named because it's a dir named "emails" by default.

but I agree `src` feels more natural without that context, which seems good.

Issue #93, opened 2022-08-05
ESM support


neuxregime:
Hello again!

I've been using [react-markdown](https://github.com/remarkjs/react-markdown) and running into issues with ESM-only-module support. For the time being I've switched to `marked` to get around these issues with CommonJS, but was wondering if the preview server is capable (or will be capable) of handling ESM libs?

Cheers

psugihara:
Hey! Yeah that seems possible. I'd guess we need to mess with the options here https://github.com/sofn-xyz/mailing/blob/main/packages/cli/src/commands/preview.ts#L53

How are you using react-markdown with your emails? I want to make sure I test for your use-case.

I'll take a look in the next few days. Thanks for bringing this up.

neuxregime:
nothing too crazy. 
```typescript
  <ReactMarkdown
        allowedElements={[
          'a',
          'li',
          'ul',
          'ol',
          'em',
          'p',
          'pre',
          'strong',
          'code',
        ]}
        linkTarget="_blank"
      >
       **bold**
</ReactMarkdown>
```

tysm!

mrlubos:
Hi @neuxregime, any chance you've got a project somewhere with instructions how to reproduce the issue?

neuxregime:
> Hi @neuxregime, any chance you've got a project somewhere with instructions how to reproduce the issue?

I'm out of town for a bit, but you can try with a simple Next.js project and put this somewhere in the welcome template

`
 <ReactMarkdown>
       **bold**
</ReactMarkdown>
`

Most of remarkjs repos are ESM only.

https://github.com/remarkjs/react-markdown

PR #118, opened 2022-08-15
swap babel and ts-node template compilation for swc

### LINKED ISSUE #102: Faster initial load
Especially in js apps (because of babel?) the initial index load of `previews.json` is very slow (like 4 seconds for the default templates.

ideas to speed this up:
1. warm-up on boot: kick off the module loading asynchronously in the preview.ts handler function
2. speedier module loading: try using swc or a different loader rather than babel
3. rust rewrite (jk)

cc @alexfarrill

psugihara:
fixes #102 

This makes the initial index request much faster (seems like 2-3x). Here's the export-previews command that had the same slow template compilation. I generated these benchmarks with hyperfine.

### TS

Before...
```
Benchmark 1: npx mailing export-previews
  Time (mean ± σ):      4.804 s ±  0.057 s    [User: 7.017 s, System: 0.649 s]
  Range (min … max):    4.720 s …  4.878 s    10 runs
```

After...
```
Benchmark 1: npx mailing export-previews
  Time (mean ± σ):      1.779 s ±  0.044 s    [User: 1.503 s, System: 0.424 s]
  Range (min … max):    1.717 s …  1.855 s    10 runs
```

### JS

Before...
```
Benchmark 1: npx mailing export-previews
  Time (mean ± σ):      2.497 s ±  0.118 s    [User: 2.290 s, System: 0.524 s]
  Range (min … max):    2.337 s …  2.712 s    10 runs
```

After...
```
Benchmark 1: npx mailing export-previews
  Time (mean ± σ):      1.704 s ±  0.042 s    [User: 1.443 s, System: 0.406 s]
  Range (min … max):    1.664 s …  1.814 s    10 runs
```


psugihara:
Fiddled with it for a minute but was not able to solve #93 in here, so I think we punt on doing that.

mrlubos (approved):

PR #122, opened 2022-08-16
fix(hotkeys): remap view hotkeys, add previous mode hotkey

mrlubos:
- refactor hotkeys to make adding more views easier (such as https://github.com/sofn-xyz/mailing/tree/gh/119-text-previews)
- add the option to jump to the previous view mode (using `[`)
- remap the option to jump to the next view mode (using `]`)
- remap the view mode keys to use more standard pattern (using `1`, `2`, `3`)

### DIFF
--- a/packages/cli/src/components/hooks/usePreviewHotkeys.tsx
+++ b/packages/cli/src/components/hooks/usePreviewHotkeys.tsx
@@ -6,36 +6,51 @@ type Options = {
   setViewMode: React.Dispatch<React.SetStateAction<ViewMode>>;
 };
 
+const hotkeysMap = {
+  showPreviews: '/',
+  viewModeDesktop: '1',
+  viewModeHTML: '3',
+  viewModeMobile: '2',
+  viewModeNext: ']',
+  viewModePrevious: '[',
+};
+
+const viewModeOrder: ViewMode[] = ['desktop', 'mobile', 'html']
+
 // Hotkeys for pages showing previews / intercepts
 // Listen for hotkeys on the document and the iframe's
 // document, so that they still work if iframe has focus.
 export default function usePreviewHotkeys({ setViewMode }: Options) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/components/hooks/usePreviewHotkeys.tsx
```diff
@@ -6,36 +6,51 @@ type Options = {
   setViewMode: React.Dispatch<React.SetStateAction<ViewMode>>;
 };
 
+export const hotkeysMap = {
```
psugihara, line 9: hell yea

psugihara (approved):
love it, thanks @mrlubos!

mrlubos:
Thanks @psugihara! Mind explaining the `Intercept.tsx` and `[previewFunction].tsx` files? I don't understand what these do, just that I needed to update the hotkeys references in them. They seem pretty much like the same file, why do we need two?

psugihara:
`[previewFunction].tsx` is for displaying the preview functions from the index. They're just rendered react components with hard-coded props.

`Intercept.tsx` is for displaying email that [the library "catches" rather than sending](https://github.com/sofn-xyz/mailing/blob/main/packages/core/src/index.ts#L93) when sendMail is called in development mode. It shows metadata about the sender. If you're developing mailing, you can see an example intercept at `/intercepts/mock`:
Screen Shot 2022-08-16 at 10 48 04 AM

Main differences:
- has all the metadata you passed into sendMail (to, from, subject, text, etc.)
- does not allow you to deliver real email yet

They used to use the iframe in different ways, but now are more similar because they both fetch json with the rendered HTML from the preview server now and pass it into the iframe with the srcDoc prop. I think having 2 files and composing shared components within them is better than combining them since they have slightly different use-cases and underlying data.

mrlubos:
Got it, thanks @psugihara!

PR #125, opened 2022-08-16
Gh/120 e2e

### LINKED ISSUE #120: move e2e tests runs folder out of the project
having the e2e_tests/runs in the project has some drawbacks:
- it magically uses node_modules in the root directory, so it's not a true test of package requirements, e.g.  moving would make it behave more like it does in the real world
- it causes redwood js install to fail and we currently have to hack around it.  moving would allow us to remove those hacks

alexfarrill:
@psugihara give this a try 
Verified by removing `@swc/core` and `@swc/register` from package.json and running, get this error as expected:
Screen Shot 2022-08-16 at 4 49 40 PM

fixes #120


scripts/e2e_test/e2e_test.rb
```diff
@@ -86,6 +101,7 @@ def build_mailing
 
     Dir.chdir(CLI_ROOT) do
       system("npx yalc add")
+      system("npx yalc publish")
```
alexfarrill, line 104: @psugihara this is very necessary.  otherwise, it works while the version number is the same, but if you update the version number it does not get added to `~/.yalc/packages/mailing`
e.g.
```
ls ~/.yalc/packages/mailing
0.4.10		0.4.11-next.0	0.6.5		0.6.8
```

scripts/e2e_test/e2e_test.rb
```diff
@@ -16,10 +16,11 @@ class TestRunner
   include TestRunnerUtils
 
   NUM_RUNS_TO_KEEP = 5
-  PROJECT_ROOT = File.expand_path(__dir__ + '/../..') 
+  PROJECT_ROOT = File.expand_path(__dir__ + '/../..')
+  TEST_ROOT = File.expand_path(PROJECT_ROOT + '/../mailing_e2e_tests')
```
psugihara, line 20: 👍

psugihara (approved):
nice. works well!

package.json
```diff
@@ -18,7 +18,7 @@
     "watch": "preconstruct watch",
     "release": "yarn build && yarn changeset publish",
     "postinstall": "preconstruct dev",
-    "init:e2e": "./scripts/init_e2e"
+    "e2e": "cd scripts/e2e_test/ && bundle && bundle exec ruby e2e_test.rb"
```
alexfarrill, line 21: :chefskiss

PR #128, opened 2022-08-17
remove hacks that were prev required for redwood apps when e2e tests …

alexfarrill:
…were in a subdirectory


psugihara (approved):
awesome

Issue #58, opened 2022-07-27
mailing render server API; production mode


psugihara:
## Context

A few users have asked for a method of running mailing as a standalone process with an HTTP API. They'd like to use it from backends that don't have access to react or can't consume nodejs packages. The new API would have one endpoint that renders a template to HTML with input props.

Thanks to @sebasalvarado, @alan-francis, and @jdotjdot for describing the use-cases in #54.

## Design

Currently the `mailing` preview server is intended only for development. That is, it's built without any consideration for handling many requests and running for a long time unattended. Are there memory leaks? How many TPS can it serve? Are there spots we can switch to async IO? Let's find out.

This design adds a "production mode" for the mailing server as well as the required API (which could also be useful for development mode features like editing props from the browser).

### CLI change

With `NODE_ENV=production` set, `mailing` should start the server in production mode, running efficiently and indefinitely.

### HTTP API change

```
GET /renders/TemplateName.json?props=JSON_PROPS
```

##### 200
``` tsx
{
  html?: string,
  errors?: object[] // prop type or template compilation errors
}
```

##### 404
```
template not found
```

## Tasks:
- [ ] cli change outlined above
- [ ] add API route above
- [ ] jest tests for API route
- [ ] manually test deploying this to fly.io
- [ ] add instructions to README for setting up and deploying a standalone mailing server

jdotjdot:
Another thought on how to deploy it: I'd be pretty fine with "production mode" for this actually being a small binary or some way to package it up as a lambda or similar, since I would imagine most production calls would take the form of "send props, get back HTML", potentially also with a validation step of some sort.  I don't necessarily need an entire server for this, I'd be fine to have this somewhere serverless.

psugihara:
> somewhere serverless

@jdotjdot do you happen to have an example of a library that you think does this well? Examples I've seen of this (e.g. [postgraphile](https://github.com/graphile/postgraphile-lambda-example) 💙) usually build out an example repo on one platform rather than a portable feature built into the library.

My default for a serverless deployment of `mailing` would be shipping a next.js app to vercel and importing a `render` function from mailng-core in an API route like we do in the demo video. Packaging as a next app might be useful because they're ubiquitous and easy to deploy on [any platform](https://nextjs.org/docs/deployment#other-services) server-full or serverless. Would an example repo doing that be helpful?

[...]

Issue #129, opened 2022-08-18
Monorepo compatibility


mrlubos:
Hey team mailing, great strides over the last few weeks towards v1!

Based on `v0.6.9`, I think the last missing piece to unlock usage in monorepos is adding support for path mappings. I have multiple packages in my project that import other packages, so my template could look like this.

```tsx
import { veryUsefulFunction } from '@repo/package';

export function MailingTemplate() {
  return (
    <Layout>
      <Button onClick={veryUsefulFunction}>
        I want to be useful!
      </Button>
    </Layout>
  );
}
```

This currently won't work with either dev or build commands. It will fail with the module not found error message. I tried to poke in mailing to see what needs to change and it seems it's the `registerRequireHooks()` method in [registerRequireHooks.ts](https://github.com/sofn-xyz/mailing/blob/main/packages/cli/src/commands/util/registerRequireHooks.ts).

I was able to manually update it to fix the module not found error, but then I ran into other issues such as:
- `SyntaxError: Unexpected token 'export'`
- `SyntaxError: Cannot use import statement outside a module`

Resolving these would require a more involved approach and it's possible this has something to do with mailing itself, so I didn't continue. Any pointers?

psugihara:
thanks @mrlubos.

a few thoughts:

1. i'm sure this is just an example but don't put onClicks in your email templates, client-side js won't work in email clients :)
2. maybe this is related to #93 
3. if you think it may be different from #93, a minimal repro example we could test fixes on would be super useful

mrlubos:
Hahah yes @psugihara, it's a href in my code. The specific use case is shared path generators between client and emails as emails deep link the application state. I will have a look at the issue and see if I can reproduce anything!

mrlubos:
@psugihara looks like it's related to #93. I'll put together a demo later

mrlubos:
@psugihara You should've received an invitation to a [CodeAnywhere repository](https://ide.codeanywhere.com/nodejs-prickly-wolf-lmenus666910/#/home/cabox/workspace/mailing). I've left the instructions to reproduce in README, it's very straightforward. Please let me know if you need any assistance with reproduction or anything else

mrlubos:
@psugihara Just checking in if you've received any emails/had time to look at this issue?

psugihara:
Hey, sorry for the delay. I got the email but haven't had time to look into the issue. When I looked at #93 I was able to repro but was not able to figure out a fix (yet).

[...]

PR #131, opened 2022-08-19
v0.7.0 Build to next app

alexfarrill:
- Introduces .mailing, remote control next app architecture
- Adds `GET /api/render` API for getting rendered HTML given props and template name
- Adds `npx mailing server`, `npx mailing server build`, `npx mailing server start`

addresses #58


alexfarrill:
What do we want this to be doing?
image

psugihara:
Nice. That's right except the manifest needs to be next to package.json I think. 

That page is expected not to work bc it's still hitting an api route that doesn't exist. We will switch it to hit /api/previews.json and that will grab info from the manifest.

alexfarrill:
Cool, that makes sense.  Pushed up a refactor to get the directory structure right

psugihara:
working on the api modules

alexfarrill:
- [ ] move render route to an API
- [ ] make `npx mailing deploy` work
- [ ] clean up scripts in package.json

psugihara:
fast :)

packages/cli/src/pages/previews/[previewClass]/[previewFunction].tsx
```diff
@@ -13,21 +19,18 @@ export const getStaticPaths = async () => {
     params: Params;
   }[] = [];
 
-  if (process.env.NEXT_PUBLIC_STATIC) {
-    const res = await fetch("http://localhost:3883/previews.json");
```
Comment (psugihara) [inline packages/cli/src/pages/previews/[previewClass]/[previewFunction].tsx:17]: removes some port hardcoding :)

psugihara:
- [ ] ensure that this setup will work when slurping up templates that are in the same directory as .mailing https://github.com/sofn-xyz/mailing/issues/129#issuecomment-1224337794

psugihara:
latest in 0.7.0-next.1

packages/cli/src/util/config.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    logPlain(`
+    ███╗   ███╗ █████╗ ██╗██╗     ██╗███╗   ██╗ ██████╗ 
+    ████╗ ████║██╔══██╗██║██║     ██║████╗  ██║██╔════╝ 
+    ██╔████╔██║███████║██║██║     ██║██╔██╗ ██║██║  ███╗
+    ██║╚██╔╝██║██╔══██║██║██║     ██║██║╚██╗██║██║   ██║
+    ██║ ╚═╝ ██║██║  ██║██║███████╗██║██║ ╚████║╚██████╔╝
+    ╚═╝     ╚═╝╚═╝  ╚═╝╚═╝╚══════╝╚═╝╚═╝  ╚═══╝ ╚═════╝ 
+    
+        `);
```
mrlubos, line 75: The important change

psugihara: needs design review from @monicakogler at some point ;)

alexfarrill: Behold!

psugihara (approved):

PR #135, opened 2022-08-23
update readme

alexfarrill:
add mailing-core to README and note:
    Note: you may want to add mailing as a dev dependency instead if you don't plan to use the preview function outside of development.  mailing-core exports buildSendMail, which returns the sendMail function


psugihara:
awesome ty

PR #148, opened 2022-08-25
Mailing telemetry

alexfarrill:


packages/core/src/util/postHog.ts
```diff
@@ -0,0 +1,16 @@
+import { PostHog } from "posthog-node";
```
alexfarrill, line 1: @psugihara do you know of a way to share this file across cli and core so I don't have to duplicate it like this?

packages/core/src/util/postHog.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const POSTHOG_API_KEY = "{POSTHOG_API_KEY}";
```
alexfarrill, line 3: @psugihara here is where we would inject the string during the build.  do I want to be looking at an swc config to do that?  I saw that swc has a transform option that sounds like maybe what i want? https://swc.rs/docs/configuration/swcrc

packages/core/src/index.ts
```diff
@@ -57,6 +58,14 @@ export function buildSendMail(options: BuildSendMailOptions) {
   }
 
   return async function sendMail(mail: ComponentMail) {
+    // anonymous telemetry
+    if (anonymousId) {
```
alexfarrill, line 62: In the core package, there's no CLI argv to specify this so I'm gonna need to get it from somewhere else.  Am I gonna be able to read mailing.config.json from the project root or do I have to think of a different way to do this?

packages/core/src/util/postHog.ts
psugihara: not sure what's best but right now we either
- duplicate files
- export from mailing-core (though i don't think this makes sense to export as a public api)

we could add a shared package or use a symlink

alexfarrill: Just gonna copy for now.  The main thing we need to decide is how we want to get the anonymousId into the sendMail call

packages/core/src/util/postHog.ts
psugihara: I think you actually want to do something with rollup/preconstruct.

swc is just used 1. internally by next.js 2. to parse templates at runtime in the CLI

alexfarrill:
todo:
- [ ] all commands should use buildHandler
- [ ] test case where mailing.config.json does not exist yet
- [ ] test case where anonymousId is null
- [ ] test case where --quiet
- [ ] capture and report errors?

packages/cli/src/util/postHog/__test__/postHog.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  afterEach(() => {
+    delete process.env.MM_DEV;
+  });
+
+  it("should call capture on the postHog client", () => {
+    const mockPostHogClient = { capture: jest.fn() };
+
+    jest
+      .spyOn(postHogClient, "postHogClient")
+      // @ts-ignore
```
alexfarrill, line 20: @psugihara if you have a sec, lmk if you have a better idea how to appease ts here

packages/cli/src/util/postHog/__test__/postHog.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    expect(mockPostHogClient.capture).toHaveBeenCalledWith({
+      distinctId: "abc",
+      event: "ate pizza",
+    });
+  });
+
+  it("should call capture on the postHog client - with config.anonymousId if options.distinctId is blank", () => {
+    // jest.spyOn(moduleManifestUtil, "getConfig").mockImplementation(() => {
```
alexfarrill, line 35: @psugihara if you have a sec, when I uncomment this, it makes this test pass but all others fail.  i think i have to clear mocks before the other tests, but that doesn't fix it.  i will take a look tomorrow unless you know what I should do

packages/cli/src/util/postHog/__test__/postHog.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    capture({
+      event: "ate pizza",
+    });
+
+    expect(mockPostHogClient.capture).not.toHaveBeenCalled();
+  });
+
+  it("should not call capture if MM_DEV", () => {
+    process.env.MM_DEV = "1";
```
alexfarrill, line 94: ⚠️

packages/core/src/index.ts
```diff
@@ -5,6 +5,14 @@ import fs from "fs-extra";
 import { render } from "./mjml";
 import { error, log } from "./log";
 import fetch from "node-fetch";
+// import postHogClient from "./util/postHog";
```
alexfarrill, line 8: we shoold revert everything in core to main for v0

packages/cli/src/util/postHog/__test__/postHog.test.ts
psugihara: think it needed mockRestore because you had changed the implementation 🤷

packages/cli/src/util/postHog/__test__/postHog.test.ts
psugihara: fixed in a different type unsafe way, but seems better 🤷

psugihara (approved):
🙌

PR #149, opened 2022-08-26
export-previews, server build cli tests

psugihara:


alexfarrill (approved):
beauty... I like the idea of using snapshot for the output.  This will work well and probably a good idea to keep everything but the runner in js.  After this gets merged I'll work on something to copy this into each e2e run and run them there, should be pretty straightforward

psugihara:
yeah i hate snapshots usually but it makes sense here where we 1) test a string we're generating and 2) it's the same every time

PR #154, opened 2022-08-29
Bundle emails

alexfarrill:
@psugihara does this also resolve your issue importing modules outside of the src directory?

This allows the frontend to get the config:

Screen Shot 2022-08-29 at 4 17 10 PM

A side effect is `npx mailing preview` no longer works, you have to use `yarn dev`

Screen Shot 2022-08-29 at 4 17 47 PM


alexfarrill:
@psugihara have another look

given that the templates bug is in main, I think we should merge this to main and release.  otherwise was gonna suggest we merge this to a release branch

image

alexfarrill:
✅ e2e tests pass

psugihara:
@alexfarrill which is the templates bug you think is in main?

psugihara:
This breaks the demo because `yarn next build` from packages/cli does not work. Could either fix that or edit this config...

Screen Shot 2022-08-31 at 4 25 09 PM

alexfarrill:
> @alexfarrill which is the templates bug you think is in main?

function assigning to templates does not have a return value here:

https://github.com/sofn-xyz/mailing/blob/fa44632cb673013fffb289781c980c88ce980694/packages/cli/src/commands/preview/server/setup.ts#L47

alexfarrill:
> 

That's so we can mark the package.json dependencies/peerDependencies as external when we esbuild - so that they don't get bundled in.  Do you think it is safe to change this to the cli package.json dependencies/peerDependencies?  It might be, I'll try it

psugihara:
o damn yeah, didn't realize that was in main

alexfarrill:
> o damn yeah, didn't realize that was in main

I am really not sure how it's working at all in main 🤦

alexfarrill:
@psugihara `yarn next build` now runs in packages/cli.  it emits to `packages/cli/.next`, is that what is intended or was it emitting to `packages/cli/src` before?  I'm not familiar enough with tsconfig to say how rootDir and outDir interact, but I'll check what it is doing in main.  lmk if you think this will work

psugihara:
yes, that should work!

psugihara:
looks like vercel enjoyed

alexfarrill:
behavior is the same in main, emits to `packages/cli/src`.  I think it probably configured rootDir automatically based on that is where all the source is, but . is more correct now that we look at ./package.json.  This seems good to me

alexfarrill:
Oh cool, I guess this is the same fix for the deploy hook that has been failing 😎

alexfarrill:
and no idea how outDir ./lib ever worked, i guess it was just ignoring that for some reason.  oh well, I think this config is good

alexfarrill:
Oh, I guess since it also has "noEmit": true ts is not emitting anything anyway and it is next doing all the emitting.  so rootDir just needs to be "." for correctness.  But then I don't know why it also started complaining about "./lib" after I changed rootDir to "." 🤷

mrlubos:
I don't even know what's going on here but it sounds like a step in the right direction? 🎉

PR #165, opened 2022-09-04
CLI test refactor

psugihara:
little one i peeled off

- ~add a test for server start~
- refactor cliExec to use async for better parallelism


psugihara:
hm, timeout error in CI

psugihara:
adding in long-polling branch

PR #166, opened 2022-09-04
Faster livereload

psugihara:
This improves the performance and reliability of previews updating in your browser when you make a change.

It refactors livereload server code into it's own module and implements a different strategy for pushing changes to the client ([long-polling](https://en.wikipedia.org/wiki/Push_technology#[4]Long_polling)). Previously there was also a bug where if you had multiple tabs open, it would not reliably reload. Switching from a boolean to vector clock fixes that.

- Long polling strategy
- Shared poll interval constants for client/server
- Adds small fetch indicator from @monicakogler for when the template is updating
- Update jest, extend timeout
- Server starts before running tests
- Feels a lot faster and more responsive ⚡️


jest.config.ts
```diff
@@ -0,0 +1,26 @@
...
+  ],
+  watchPathIgnorePatterns: [
+    "<rootDir>/emails/",
+    "<rootDir>/packages/cli/src/emails/",
+    "<rootDir>/packages/cli/src/generator_templates/ts/emails/",
+    "<rootDir>/.mailing/",
+    "tmp-testMailQueue.json",
+  ],
+  setupFilesAfterEnv: ["<rootDir>/testSetup.ts"],
+  // TODO: keep testTimeout low and use jest.setTimeout for long ones
```
psugihara, line 19: couldn't get it to work this way, not sure why

.github/workflows/node.js.yml
```diff
@@ -5,9 +5,6 @@ name: Node.js CI
 
 on:
   push:
-    branches: [main]
-  pull_request:
-    branches: [main]
```
psugihara, line 10: github actions are free for open source, so lets run tests on every push

packages/cli/src/commands/preview/server/setup.ts
```diff
@@ -4,17 +4,18 @@ import pkg from "../../../../package.json";
...
   copy,
   mkdir,
-  mkdirp,
   readdir,
   remove,
   rm,
   writeFile,
   readFile,
   appendFile,
+  copyFile,
```
alexfarrill, line 13: not used

alexfarrill (changes requested):
seems to need a few reloads before it starts working? https://www.loom.com/share/b227e6ba45414fa3b63cfb688c1cd65d

psugihara:
excellent find @alexfarrill, thx for testing.

I had removed the initial polling call when debugging so it would be LONG_POLLING_INTERVAL (20 seconds) until it kicked in. It now fires a long poll on mount.

psugihara:
Haven't figured out a good way to write a good test for the front-end of this yet clearly :)

alexfarrill (approved):
working great now!

PR #167, opened 2022-09-05
New templates in readme

monicakogler:


packages/cli/README.md
```diff
@@ -97,14 +102,14 @@ const transport = nodemailer.createTransport({
 
 ```tsx
 import { sendMail } from "emails";
-import Welcome from "emails/Welcome";
+import Welcome from "emails/AccountCreated";
```
psugihara, line 105: should be `import AccountCreated from "emails/AccountCreated";`

packages/cli/README.md
```diff
@@ -114,49 +119,29 @@ sendMail({
...
+For example, here's `emails/previews/AccountCreated.tsx`:
 
 ```tsx
-import Welcome from "../Welcome";
+import AccountCreated from "../AccountCreated";
 
-export function toAmelita() {
-  return <Welcome name="Amelita" />;
+export function accountCreated() {
+  return <accountCreated name="Amelita" />;
```
psugihara, line 128: should be `<AccountCreated`

psugihara (changes requested):
noticed a couple things

PR #173, opened 2022-09-06
Refactor lodash templates

alexfarrill:
@psugihara let me know if you like this idea, it makes it a bit easier to see what is going on in the dynamic files


psugihara (commented):
Seems cool. Is there a way to get syntax highlighting? Maybe `.ejs` syntax would be better in that regard?

PR #174, opened 2022-09-07
sendMail telemetry in core

### LINKED ISSUE #169: Telemetry: add telemetry to sendMail
- [x] should be a noop when NODE_ENV !== "production" ?

psugihara:
Adds simple telemetry to sendMail.

- should send if capture fails
- should send if anonymousId is not found

fixes #169 

TODO: 
- [x] manually verify this works in an app by adding it to a next app and calling sendMail from an api route like the demo video


psugihara:
ok it works but you must flush after the call apparently
Screen Shot 2022-09-07 at 11 09 31 AM

Screen Shot 2022-09-07 at 11 21 44 AM

alexfarrill (approved):

PR #177, opened 2022-09-08
fix #163 - esbuild node_modules excludes

### LINKED ISSUE #163: Preview server error since 0.7.2 in yarn workspace
Have tried to update to `mailing@^0.7.2` and unfortunately I'm getting this error when attempting to run `mailing preview`.

```
Server Error
Error: Cannot find module 'http'

This error happened while generating the page. Any console logs will be displayed in the terminal window.
Call Stack
webpackEmptyContext
file:///Users/willviles/Code/@Redbrain/packages/email/.mailing/.next/server/pages/index.js (22:10)
../../node_modules/axios/lib/adapters/http.js
webpack-internal:///./src/moduleManifest.js (1924:20)
__require2
webpack-internal:///./src/moduleManifest.js (34:56)
getDefaultAdapter
webpack-internal:///./src/moduleManifest.js (2240:27)
../../node_modules/axios/lib/defaults/index.js
webpack-internal:///./src/moduleManifest.js (2259:22)
__require2
webpack-internal:///./src/moduleManifest.js (34:56)
../../node_modules/axios/lib/core/transformData.js
webpack-internal:///./src/moduleManifest.js (2338:24)
__require2
webpack-internal:///./src/moduleManifest.js (34:56)
../../node_modules/axios/lib/core/dispatchRequest.js
webpack-internal:///./src/moduleManifest.js (2362:29)
__require2
webpack-internal:///./src/moduleManifest.js (34:56)
```

There are no errors in a vanilla project, however I'm using Mailing within a yarn workspace, which already contains a next.js app. 

Pinning to `mailing@0.7.1` and adding a the following nohoist rules to my root `package.json` ensures the error disappears:

```json
{
  "workspaces": {
    "nohoist": [
      "**/mailing",
      "**/mailing-core"
    ]
  }
}
```

alexfarrill:
This fixes the problem in https://github.com/willviles/mailing-yarn-workspaces

### DIFF
--- a/packages/cli/src/commands/preview/server/__test__/setup.test.ts
+++ b/packages/cli/src/commands/preview/server/__test__/setup.test.ts
@@ -3,7 +3,6 @@ import path from "path";
 import {
   packageJsonVersionsMatch,
   bootstrapMailingDir,
-  getNodeModulesDirs,
   COMPONENT_FILE_REGEXP,
 } from "../setup";
 import childProcess from "child_process";
@@ -117,18 +116,6 @@ describe("setup", () => {
 
       await bootstrapMailingDir();
       expect(fsExtra.copy).not.toHaveBeenCalled();
-    });
-  });
-
-  describe("getNodeModulesDirs", () => {
-    it("should return an array with all the node_modules folders", () => {
-      jest.spyOn(path, "resolve").mockImplementation(() => "/a/b/c");
-      expect(getNodeModulesDirs()).toEqual([
-        "../../../node_modules",
-        "../../node_modules",
-        "../node_modules",
-        "./node_modules",
-      ]);
     });
   });
 
--- a/packages/cli/src/commands/preview/server/__test__/setup.test.ts
+++ b/packages/cli/src/commands/preview/server/__test__/setup.test.ts
@@ -1,5 +1,4 @@
 import fsExtra from "fs-extra";
-import path from "path";
 import {
   packageJsonVersionsMatch,
   bootstrapMailingDir,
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/preview/server/setup.ts
```diff
@@ -200,8 +199,11 @@ async function buildManifest(
     format: "esm",
     jsx: "preserve",
     external: [
-      ...Object.keys(pkg.dependencies || {}),
-      ...Object.keys(pkg.peerDependencies || {}),
+      "./node_modules/*",
```
alexfarrill, line 202: @psugihara this is a better way to do this.  any thoughts on best way to do this programmatically?  naive solution I could just prepend "../" for however deep cwd is, but there is probably a better way.  is there something you can check in `process`? i.e. ask node/npm, etc?

psugihara: this seems nice and simple

I think use `process.cwd().split("/").length` and just generate the string array like that.

This plugin looks in the filesystem but this way seems quicker if the string array works https://github.com/pradel/esbuild-node-externals/blob/main/esbuild-node-externals/src/utils.ts#L23

packages/cli/src/commands/preview/server/__test__/setup.test.ts
```diff
@@ -25,101 +25,107 @@ function mockReadJSONSyncToThrowErrno() {
   });
 }
 
-describe("packageJsonVersionsMatch", () => {
-  it("should return false if .mailing package.json does not exist", async () => {
-    mockReadJSONSyncToThrowErrno();
-    const match = await packageJsonVersionsMatch();
-    expect(match).toBe(false);
+describe("setup", () => {
+  beforeEach(() => {
```
alexfarrill, line 29: this 3 liens is the only change that's not formatting

alexfarrill:
e2e tests ✅

psugihara (approved):
looks great, nice fix!

psugihara (approved):

PR #191, opened 2022-09-12
Gh/189 minify

alexfarrill:


psugihara (approved):
nice work. so sick that you can just pass the option.

PR #194, opened 2022-09-15
Gh/193 iframe restyle

alexfarrill:
Last thing: the figma has black icons for active state and white non-active. Is there a trick to export the svg from figma in a way that I can use css "fill" to change the color?


psugihara:
@alexfarrill idk but i kind of doubt it

@monicakogler do you know?

monicakogler:
unless i'm misunderstanding, this should work whether you do figma's "copy as svg" or export as svg. either way, you can just edit the fill in the code.

Screen Shot 2022-09-15 at 8 51 50 AM

Screen Shot 2022-09-15 at 8 52 23 AM

monicakogler:
lgtm other than the deselected icon color you mentioned (should be white). this'll break the hover state (white on white), but i just pushed a `gray-700` that we can use for the bg with the white icon. 

also the border renders diff irl than in figma – if it's easy can we make it transparent in the active state please? i.e. transprent border around the blue

Screen Shot 2022-09-15 at 8 57 36 AM

psugihara:
If you want to use those in your jsx, @alexfarrill I think you can do something like this: https://tailwindcss.com/docs/configuration#referencing-in-java-script

PR #199, opened 2022-09-15
type sendMail return value

### LINKED ISSUE #164: sendMail has `any` result type
The result of the `sendMail` call is currently typed as `any`. It'd be helpful if it was strongly typed.

```ts
  const ownerEmailResult = await sendMail({
    subject: 'Here's an email!',
    to: owner.body?.email,
    component: (
      <EventCreatedOwnerTemplate {...req.body} name={owner.body?.full_name!} />
    ),
    forceDeliver: true,
  });

  // ownerEmailResult is type <any>
```

alexfarrill:
fixes #164

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -72,7 +72,7 @@ export function buildSendMail(options: BuildSendMailOptions) {
     }
   }
 
-  return async function sendMail(mail: ComponentMail) {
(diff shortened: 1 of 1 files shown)


psugihara:
this is already the inferred type so I don't think this fixes the issue

iduuck:
Yeah, that does not fix the issue.

PR #201, opened 2022-09-16
v0.8.0: Immersion UI

psugihara:
Almost there with the new UI ::)

- Navigate through templates quickly
- See what preview text might look like in an email client
- New default templates set a better example for making focused templates vs catch-alls like the deprecated TextEmail

known todos before merging:
- [x] remove "/" hotkey that goes to index #202
- [x] add "command+period" hotkey to toggle to full screen iframe #202
- [x] mobile view #202
- [x] qa
- [x] e2e tests green

Screen Shot 2022-09-15 at 8 55 46 PM
Screen Shot 2022-09-15 at 8 56 20 PM


packages/cli/src/commands/preview/server/setup.ts
```diff
@@ -18,6 +18,8 @@ import { template } from "lodash";
 import { getNodeModulesDirsFrom } from "../../util/getNodeModulesDirsFrom";
 
 export const COMPONENT_FILE_REGEXP = /^[^\s-]+\.[tj]sx$/; // no spaces, .jsx or .tsx
+export const DOT_MAILING_IGNORE_REGEXP =
+  /__test__|generator_templates|yarn-error.log|src\/commands|src\/index\.ts$|src\/dev\.js$|\.mailing$|\.next|node_modules$|^cypress/;
```
alexfarrill, line 22: @psugihara was there a reason we added `src\/commands` in this branch?  The e2e tests fail because the manifest and feManifest ejs templates are in a subdir of src\/commands

psugihara: nah, i don't think there is a reason

psugihara:
mailing@0.8.0-next.0
mailing-core@0.8.0-next.0

psugihara:
OK, I think the latest mailing@next beta is good to go.

Will let it bake for a minute if anyone else has a minute to try!

- [ ] @alexfarrill ✅

alexfarrill:
Nice, will try it in a moment

monicakogler:
fantastic! lgtm.

only thing i noticed: on mobile, the header padding is asymmetric and the title is off-center, but not blockers imo. i'll fix in a followup.
<br/>

Screen Shot 2022-09-19 at 12 10 05 PM

alexfarrill:
@psugihara let's merge and release so that the fix for #206 also goes out

PR #205, opened 2022-09-18
improve hotkey menu design

psugihara:
before
Screen Shot 2022-09-18 at 1 05 01 PM

after
Screen Shot 2022-09-18 at 1 04 39 PM


psugihara:
Screen Shot 2022-09-18 at 11 28 54 PM

Issue #206, opened 2022-09-19
Error when exporting with minify true


im-sticky:
First off I would like to say how much I appreciate the fast turnaround on the export minify option. However I seem to be running into an error when it is enabled, both on my project as well as the examples that get created when running `mailing` for the first time.

The error is as follows:
```
mailing Exporting minified preview html to
mailing ./previews_html
mailing   |-- text_email_reset_password.html
mailing   |-- welcome_to_amelita.html
mailing ✅ Processed 2 previews
/dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:1928
                          if (errors.length) {
                                     ^

TypeError: Cannot read properties of undefined (reading 'length')
    at /dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:1928:38
    at Array.map (<anonymous>)
    at _callee$ (/dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:1919:78)
    at tryCatch (/dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:141:17)
    at Generator._invoke (/dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:121:24)
    at Generator.next (/dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:172:21)
    at asyncGeneratorStep (/dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:405:24)
    at _next (/dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:427:9)
    at /dev-testing/mailing-demo/node_modules/mailing/dist/mailing.cjs.dev.js:434:7
    at new Promise (<anonymous>)
```

Despite saying at least those 2 of the 4 previews are processed there is not HTML output.

package.json
```
{
  "dependencies": {
    "mailing": "^0.7.9",
    "mailing-core": "^0.7.9",
    "next": "^12.3.0",
    "react": "^18.2.0",
    "react-dom": "^18.2.0"
  }
}
```

mailing.config.json
```
{
  "typescript": false,
  "emailsDir": "./emails",
  "outDir": "./previews_html",
  "anonymousId": null,
  "minify": true
}
```

Code aside from minify option is not altered in any way after running initial `mailing` command.

alexfarrill:
[...]

Issue #144, opened 2022-08-25
Add e2e test for `mailing export-previews`


psugihara:
Think we shipped a bugged version in 0.7.0 (see #142). Let's catch in tests next time.

alexfarrill:
Brainstorming some ideas for how this could work:
## Option 1: jest tests
After cypress tests run, we cd into the current run, yarn install jest, copy jest tests in and run them.  The jest tests run the CLI code and show the output.  We already have yarn so setup will be pretty easy, but it might be a pain to write tests for the CLI commands and capture the output of the subprocesses with node.

## Option 2: ruby minitest tests
Same idea but with ruby -- copy in a Gemfile with bundler and minitest and a bunch of tests that we keep in the e2e_tests/ruby/unit folder in, bundle install and then run the tests.  handling the io for the subprocesses will probably be easier?

psugihara:
I think this works pretty well (would have caught one of the bugs in #142). I'll try on server commands.

<img width="576" alt="Screen Shot 2022-08-25 at 9 37 53 PM" src="https://user-images.githubusercontent.com/282016/186823218-47fac677-9e33-4a88-9978-22fcf9d4d0c4.png">

https://github.com/sofn-xyz/mailing/pull/149

psugihara:
Hm, but the other `server build` error was related to not having types in the production dep list, so this wouldn't catch.

PR #207, opened 2022-09-19
add embedded jest test for export-previews

alexfarrill:
Repros #206 

Screen Shot 2022-09-19 at 12 44 49 PM


alexfarrill:
@psugihara you will want to `rm -rf scripts/e2e_test/cache` before running since this adds jest to the template projects' package.json

alexfarrill:
verified that the minify refactor fixes #206

psugihara (approved):
awesome. this seems like a good direction for #144

PR #210, opened 2022-09-21
Homepage 2

monicakogler:
this is close. need to test on actual mobile device, and was just thinking about the bullets in the blue section with @psugihara. considering making "serverless self-hosting" its own section...


alexfarrill:
Looks good, two questions:
- Should we start with the hero image visible?  I think seeing the product right away gives a lot of information
- I think we should swap the order of the social proof with the product details below it, I want some more product details before looking for social validation.  wdyt?

monicakogler:
@alexfarrill cool i made the hero image visible on load. makes sense about testimonials, will update with the other change.

psugihara (approved):
think this is feeling goood :)

monicakogler:
@psugihara will u sanity check me on dark mode please?

psugihara (approved):
look great to me

PR #211, opened 2022-09-21
fix mailing server build

psugihara:
Think I broke `mailing server build` (noticed while trying to figure out `mailing deploy`).

This fixes for me.


psugihara:
seems eminently safe. releasing.

alexfarrill:
weird, i will try to think of a way to add the server check to the ci

PR #212, opened 2022-09-21
jest tests for server commands

### LINKED ISSUE #208: e2e: add additional cli command jest tests
- [ ] mailing server
- [ ] mailing server start
- [x] mailing server build

alexfarrill:
fixes #208


package.json
```diff
@@ -21,7 +21,8 @@
     "watch": "preconstruct watch",
     "release": "yarn build && ./scripts/prepare-release && yarn changeset publish",
     "postinstall": "preconstruct dev",
-    "e2e": "cd scripts/e2e_test/ && bundle && bundle exec ruby e2e_test.rb"
+    "e2e": "cd scripts/e2e_test/ && bundle && bundle exec ruby e2e_test.rb",
+    "e2e:jest": "yarn jest --rootDir=scripts/e2e_test/jest_tests -c scripts/e2e_test/jest_tests/jest.config.js"
```
alexfarrill, line 25: use this to update snapshots in dev, etc.

psugihara (approved):
awesome, this would have caught that bug

alexfarrill:
Do you think we need to test the other server commands?  I'm not sure how we'd check start unless we capture the process and kill it or something.  But I figure build will probably capture 99% of the same thing? wdyt?

psugihara:
yeah i think this is the 99% case... if this works start is very likely to work.

I don't think it would be too hard to start the server and kill it if we want to do that in the future though (https://nodejs.org/api/child_process.html#subprocesskillsignal). Another way to do it would be with `start-server-and-test` like in CI, but rather than running dev.js, it would run `mailing server`.

PR #222, opened 2022-09-27
Use @faire/mjml-react react bindings

IanEdington:
## Describe your changes
Use @faire/mjml-react for mjml\<\>react bindings rather than the unmaintained mjml-react.

## Issue link
#203

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.

I've tested it locally by running `yarn dev` and browsing the email templates. This is a very nice tool 🙏 

Most of the tests are passing. 1 test is breaking locally for me. The same test is failing on the main branch locally. Hopefully it passes CI.

* livereload › returns pollers immediatly when vectorClock=0

PR #224, opened 2022-09-28
Use @faire/mjml-react react bindings

IanEdington:
a second attempt at: https://github.com/sofn-xyz/mailing/pull/222

`yarn test` is passing (asside from livereload › returns pollers immediatly when vectorClock=0)

The e2e tests passed once `yarn e2e` but I wasn't able to get them to pass just before pushing. Would someone be able to run the e2e and let me know if there are any issues? I'm not confident I have my machine set up correctly

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@babel/parser": "^7.18.11",
+    "@faire/mjml-react": "^2.1.2",
     "@reecelucas/react-use-hotkeys": "^1.3.5",
     "@swc/core": "^1.2.229",
     "@swc/register": "^0.1.10",
@@ -49,7 +50,6 @@
     "@types/jsdom": "^20.0.0",
     "@types/lodash": "^4.14.184",
     "@types/mjml": "^4.7.0",
-    "@types/mjml-react": "^2.0.4",
     "@types/node": "^17.0.24",
     "@types/node-fetch": "^2.6.1",
     "@types/nodemailer": "^6.4.4",
@@ -68,7 +68,6 @@
     "lodash": "^4.17.21",
     "mailing-core": "^0.8.2",
     "mjml": "^4.12.0",
-    "mjml-react": "^2.0.7",
     "node-fetch": "^2.6.7",
     "nodemailer": "^6.7.5",
     "open": "^8.4.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@babel/parser": "^7.18.11",
-    "@faire/mjml-react": "^2.1.2",
+    "@faire/mjml-react": "^2.1.4-2",
     "@reecelucas/react-use-hotkeys": "^1.3.5",
     "@swc/core": "^1.2.229",
     "@swc/register": "^0.1.10",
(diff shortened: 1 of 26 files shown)


packages/cli/package.json
```diff
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@babel/parser": "^7.18.11",
+    "@faire/mjml-react": "^2.1.4-2",
```
IanEdington, line 43: I'll change this to `2.1.4` before merging

IanEdington:
Ok this morning I'm getting this pretty consistently

```
Module not found: Can't resolve 'encoding' in '/Users/ian/code/faire/mailing_e2e_tests/runs/20220928213816/standalone/node_modules/node-fetch/lib'
error - Error: ENOENT: no such file or directory, open '(api)/../node_modules/uglify-js/lib/utils.js'
    at Object.openSync (node:fs:599:3)
    at Object.readFileSync (node:fs:467:35)
    at eval (webpack-internal:///(api)/../node_modules/uglify-js/tools/node.js:20:19)
    at Array.map (<anonymous>)
    at eval (webpack-internal:///(api)/../node_modules/uglify-js/tools/node.js:19:30)
    at eval (webpack-internal:///(api)/../node_modules/uglify-js/tools/node.js:24:2)
    at Object.call (/Users/ian/code/faire/mailing_e2e_tests/runs/20220928213816/standalone/.mailing/.next/server/pages/api/previews.js:9613:1)
    at __webpack_require__ (/Users/ian/code/faire/mailing_e2e_tests/runs/20220928213816/standalone/.mailing/.next/server/webpack-api-runtime.js:33:43)
    at eval (webpack-internal:///(api)/../node_modules/html-minifier/src/htmlminifier.js:8:16)
    at Object.(api)/../node_modules/html-minifier/src/htmlminifier.js (/Users/ian/code/faire/mailing_e2e_tests/runs/20220928213816/standalone/.mailing/.next/server/pages/api/previews.js:3909:1) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '(api)/../node_modules/uglify-js/lib/utils.js'
}
```

IanEdington:
Thanks for looking into this. I suspect it an issue with uglify-js being imported into a browser context because of this line in the mjml-react index.js file:

https://github.com/Faire/mjml-react/blob/d36f98ad8e57826c29491389c9816f7c4d06a6ff/src/index.js#L3
```
import { minify as htmlMinify } from 'html-minifier';
```

In our app we had to write a noop mock for `uglify-js` to get around this.

We are going to start focusing on releasing a v3 of mjml-react and part of that will be splitting out the render function, and other non core functions, out of the `index.js` file. This should hopefully solve your problem.

I'll close this ticket and try again once v3 is released

psugihara:
Sounds good, thanks @IanEdington.

I don't think it would be imported into a browser context (we never import anything mjml related in client js), but Next does a lot of bundling magic even for api routes in order to ship them to serverless functions.

I think that minifier line is in the mjml-react version we're currently using  so I kind of think that's a red herring 

That said, if I knew for sure what the problem was I'd have fixed it.

IanEdington:
thanks, makes sense. My previous comment was based on solving a similar issue when using mjml-react in a browser context. Could definitely be a similar issue with a different root cause

PR #230, opened 2022-10-07
Cypress in CI

elliothursh:
## Describe your changes
Updates .tool-versions of node and yarn to latest versions. Node 16 is moving to maintenance mode starting next week. Not sure how this works with vercel deploys, i'll revert the upgrade if it's a pain.

Adds a cypress test action to run cypress tests for CLI.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18
-          cache: 'yarn'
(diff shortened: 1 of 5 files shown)


alexfarrill:
Nice!  Giving it a go...

alexfarrill (approved):
- tools installed
- jest tests passed ✅ 
- e2e tests passed ✅ (the redwood ones currently don't pass in main, so ignore those)

i got prompted to install a more recent vercel ci, but that seemed to work too

PR #231, opened 2022-10-07
Move default community files to .github repo

elliothursh:
## Describe your changes

Just a little house keeping, I moved these files into the .github repo so that mailing-stand-alone app would also be able to use them. 


psugihara (approved):
cool, didn't know you could do that

PR #235, opened 2022-10-07
Share tailwind theme

elliothursh:
## Describe your changes
Refactor our tailwind configs to share a single them.

## Issue link

N/A


psugihara:
Yeah idk the best way to share code between the packages. One way would be with a shared npm module but seems like a bunch of overhead. A probably easy but hacky option would be using a symlink.

PR #242, opened 2022-10-09
Fix web type error

psugihara:
#241 broke web deploys. This fixes the type error by switching the const back to var so that you can assign it but I'm not sure what the other problem is @elliothursh


elliothursh:
Oops! Looks like prettier messed up that the first-post.mdx file.

psugihara:
nice, ty

Issue #233, opened 2022-10-07
Analytics Proposal?


elliothursh:
## A clear and concise description of what you want to happen.
Add click tracking metrics by integrating directly with an analytics provider of your choosing (Posthog/Amplitude/Segment)

To do this we'd need to:
- Replace all URLs in emails with Mailing proxy urls
- Build proxy endpoint that triggers an event to Analytics provider and redirects user to original URL.

As a user, you should be able to choose which analytics provider you want to use (start with posthog for now?)

## Add any other context or screenshots about the feature request here.
Related to #229

PR #248, opened 2022-10-10
Analytics Providers

elliothursh:
## Describe your changes
This builds out initial analytics tracking for click and open tracking. It allows a user to send events to various analytics providers by setting the appropriate environment variables. 

Initially has integrations with Posthog, Axiom. I need to figure out how to actually set these environment variables during a deployment. 

## Issue link
#233


packages/cli/src/pages/api/messages/index.ts
```diff
@@ -0,0 +1,29 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import createMessage from "../../../util/createMessage";
+
+type Data = {
+  error?: string;
+  message?: any;
+};
+
+export default async function handler(
```
psugihara, line 9: I think we need to check the API key in here

packages/cli/src/pages/api/sendMail.ts
```diff
@@ -37,75 +20,56 @@ async function validApiKey(apiKey: string | string[] | undefined) {
...
-    references,
-  } = req.body;
+  const { templateName, previewName, props, ...mailOptions } = req.body;
+
+  let html = req.body.html;
 
   if (!(await validApiKey(req.headers && req.headers["x-api-key"])))
     return res.status(401).json({
       error: "invalid API key",
     });
```
psugihara, line 35: other API key check is here

packages/cli/src/pages/api/sendMail.ts
```diff
@@ -1,111 +1,64 @@
...
-  return render(React.createElement(Template as any, props as any));
-}
-
-async function validApiKey(apiKey: string | string[] | undefined) {
-  if (!apiKey) return false;
-
-  const host = process.env.MM_DEV ? "http://localhost:3883" : "yourInstallUrl";
-  const response = await fetch(`${host}/api/apiKeys/${apiKey}/validate`);
-
-  return 200 === response.status;
```
psugihara, line 35: @alexfarrill I forget... did we want this to be a network call to get potential caching benefits? I switched to hitting the DB directly but then thought maybe that's why we did it. I _think_ just hitting the DB should be OK because we're already going to be hitting it in anything that requires an API key.

psugihara: ^ discussed w alex. hitting db directly seems good. to optimize later we can cache/hardcode api keys at build time for instant lookup.

packages/cli/prisma/schema.prisma
```diff
@@ -39,3 +39,56 @@ model ApiKey {
...
+
+model MessageContent {
+  // Note: This table will likely grow large over time, we will likely
+  // want to truncate it in the future to only retain last N messages.
+  // We also might need to consider moving it to an external data store like S3.
+  id        String   @id @default(cuid())
+  createdAt DateTime @default(now())
+  html      String?
+  messageId String
+  Message   Message  @relation(fields: [messageId], references: [id])
```
elliothursh, line 75: @psugihara @alexfarrill Just a heads up, I switched where the relation is stored for this. In the originally agreed upon schema the relation was stored on message (well technically Send). I flipped it so that if/when we truncate the messagecontent table, we'll be able to keep referential integrity. Does that make sense?

psugihara: 👍 yep seems good

packages/core/src/index.ts
```diff
@@ -72,44 +77,49 @@ export function buildSendMail<T>(options: BuildSendMailOptions<T>) {
...
+
+    // Get html from the rendered component if not provided
+    let derivedTemplateName;
+    if (component && !mailOptions.html) {
+      const { html: renderedHtml, errors } = render(component);
+      if (errors?.length) {
+        error(errors);
+        throw new Error(errors.join(";"));
+      }
+      derivedTemplateName = component.type.name;
```
psugihara, line 113: Hmmm. This is why we were going to lint.

Still feels sketchy because we re;y on the filename for all of the template naming in mailing cli (e.g. for the file tree and template names in the moduleManifest), but here we use the function name which can be different. Trying to figure out if there's a better way to do this. Ideally we'd use the filename here somehow.

packages/core/src/index.ts
```diff
@@ -135,14 +145,58 @@ export function buildSendMail<T>(options: BuildSendMailOptions<T>) {
...
+        const stringHtml = mailOptions.html?.toString();
+        if (stringHtml) {
+          mailOptions.html = instrumentHtml({
+            html: stringHtml,
+            messageId: messageId,
+            apiUrl: MAILING_API_KEY,
+          });
+        }
+      } else {
+        const json = await hookResponse.json();
```
psugihara, line 179: should decode json in a try/catch in case it's a 500 and it throws a parsing error

psugihara (changes requested):

packages/core/src/index.ts
psugihara: re-opened #284

psugihara: https://github.com/sofn-xyz/mailing/pull/287

packages/core/src/index.ts
```diff
@@ -72,44 +77,49 @@ export function buildSendMail<T>(options: BuildSendMailOptions<T>) {
...
+      templateName,
+      previewName,
+      forcePreview,
+      ...mailOptions
+    } = mail;
+    mailOptions as SendMailOptions;
+    mailOptions.from ||= options.defaultFrom;
+
+    const previewMode =
+      !options.forceNotTestMode &&
```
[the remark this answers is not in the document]
elliothursh: opted to call dangerouslyForceDeliver in tests, and added a second clause to the testMode branch in sendmail.

psugihara (approved):
cool thx for changes

dangerouslyForceDeliver: now with even more danger

PR #254, opened 2022-10-13
sendMail API

alexfarrill:


packages/core/src/index.ts
```diff
@@ -41,15 +37,18 @@ export async function clearTestMailQueue() {
...
     throw e;
   }
 }
 
 export function buildSendMail(options: BuildSendMailOptions) {
-  const testMode = process.env.TEST || process.env.NODE_ENV === "test";
+  const testMode =
+    process.env.TEST ||
+    process.env.NODE_ENV === "test" ||
+    process.env.MAILING_CI;
```
alexfarrill, line 51: @psugihara the reason this doesn't work is we do `import { buildSendMail } from "mailing-core"` (the node_modules package) in the generator templates.  So a change to core will not be reflected when in dev mode.  Do you think we should:
a) cherry-pick this into mailing-core and release it? seems harmless
b) get the generator templates to reference packages/core and search/replace them to the current line during build process

psugihara: Doing A seems better if we have to do one of those. I would have thought that pre-construct would handle yalcing mailing-core though.

alexfarrill: true ```>mailing ls -l node_modules/mailing-core
lrwxr-xr-x  1 alexfarrill  staff  16 Oct 12 13:57 node_modules/mailing-core -> ../packages/core``` maybe I just have to yarn build it before?  will try that

alexfarrill: OK that worked, I guess you have to build because it is using the code in the dist folder

alexfarrill: I get it, in core's package.json we tell it that main is `"dist/mailing-core.cjs.js",` so that's where it looks when it does module resolution

testUtilIntegration.ts
```diff
@@ -0,0 +1,42 @@
...
+const WEB_TABLE_NAMES = ["NewsletterSubscriber"];
+
+const CLI_TABLE_NAMES = ["ApiKey", "Organization", "User"];
+
+export function truncateCliDatabase() {
+  const truncateSql = CLI_TABLE_NAMES.map(
+    (tableName) => `TRUNCATE TABLE "${tableName}" CASCADE;`
+  ).join(" ");
+  debug("Running TRUNCATE on ", process.env.DATABASE_URL_TEST);
+  execSync(`echo '${truncateSql}' | psql ${process.env.DATABASE_URL_TEST}`, {
```
alexfarrill, line 28: @psugihara @elliothursh the tests are flakey with this strategy because TRUNCATE does some asynchronous stuff inside postgres.  Adding a sleep here resolves it, but before I go there do you know if there is a way to check if the command has been committed (like checking an oplog or something) or a way to make it more like a synchronous command like wrapping it in TRANSACTION or something like that?

psugihara: Hm, not sure. An alternative could be running all the truncates then polling each table for 0 count in a loop.

alexfarrill (commented):
@psugihara I like this better because no fake password and it will just 404 (and log in debug) if you aren't using sessions.  wdyt?

packages/cli/tailwind.config.js
```diff
@@ -15,6 +15,9 @@ module.exports = {
         "gray-500": "#555",
         "gray-600": "#444",
         "gray-700": "#333",
+        mint: "#C3F2BC",
```
alexfarrill, line 18: @psugihara is this naming ok these 3 lines?

psugihara: I think let's try to match tailwind color names. cc @monicakogler who did those and is trying to keep them in sync w Figma colors.

alexfarrill: tia @monicakogler , these are the button colors on the Account page

psugihara:
@alexfarrill that session scheme seems good. Maybe we render something helpful if you have a MAILING_DATABASE_URL set but no session password set.

packages/cli/tailwind.config.js
monicakogler: cool, assuming this is for the button. the text is `green-300` and the border is `emerald-700`. hopefully it'll be easier now to know the colors from figma

alexfarrill: Thanks

packages/cli/package.json
```diff
@@ -86,6 +94,8 @@
     "yargs": "^17.5.1"
   },
   "scripts": {
+    "dev": "rm -rf .mailing && yarn prisma migrate dev && src/dev.js",
+    "ci:server": "NODE_ENV=test rm -rf .mailing && yarn prisma generate && yarn prisma migrate deploy && src/dev.js > test.log",
     "ci:server:nohup": "rm -rf .mailing && nohup src/dev.js --quiet > /dev/null 2>&1 &"
```
elliothursh, line 99: does this need to be updated to include yarn prisma generate?

packages/cli/src/pages/settings.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <div className="col-span-3">
+                <p>Welcome back {props.user?.email}</p>
+              </div>
+
+              <div className="mt-16 col-span-3"></div>
+              <h2 className="col-span-2 text-3xl">API Keys</h2>
+              <div className="col-span-1 text-right">
+                <button
+                  onClick={createApiKey}
+                  className="text-sm text-green-300 border-emerald-700 border-[1px] rounded-lg p-[13px]"
```
elliothursh, line 75: border-px

elliothursh (approved):
Cool 👍

packages/cli/package.json
elliothursh: the answer is no.

PR #256, opened 2022-10-14
unify cli and web tailwind config colors with figma colors

monicakogler:
Screen Shot 2022-10-14 at 10 36 39 AM
Screen Shot 2022-10-14 at 10 37 02 AM


psugihara (changes requested):
I think this is caused by the blue key, not sure:

Screen Shot 2022-10-14 at 11 08 27 AM

monicakogler:
woops thanks. going with the alt syntax rather than replacing all instances of `blue`

psugihara (approved):
looks good!

packages/web/tailwind.config.js
```diff
@@ -7,20 +7,28 @@ module.exports = {
   theme: {
```
alexfarrill, line 7: @psugihara we should be able to symlink this, no?

psugihara: sure i'll try

PR #258, opened 2022-10-14
fix link:emails

psugihara:
## Describe your changes

I had switched vercel build settings to deploy more like users deploy, but forgot to fix this command in main. Works now.

Screen Shot 2022-10-14 at 3 57 24 PM


alexfarrill:
:nice

Issue #245, opened 2022-10-10
Safe delivery from development mode


psugihara:
Usually in development mode we intercept emails and don't send them. Sometimes you do want to deliver from development mode for testing purposes (e.g. sending to Litmus or yourself for QA). Today you can get around that by passing the `forceDeliver` flag to `sendMail`, but that is pretty dangerous and not a very findable feature.

Let's ship 2 changes to mitigate this:

1. Change the name of `forceDeliver` to `dangerouslyForceDeliver` so that it is taken very seriously. I'd rather not pipe this type of delivery though the preview server because it is useful for testing what happens in prod.

2. [needs design] Add a button to the intercept screen (pictured) that allows you to actually deliver to recipients (or just a single email?). This should include a confirmation flow with the text **_"This email will be sent to 1,427 people. Are you sure you want to deliver?"_**. My hope is 99% of the deliver-from-dev use-cases will use this.

<img width="1395" alt="Screen Shot 2022-10-10 at 9 18 11 AM" src="https://user-images.githubusercontent.com/282016/194911110-ad72cb8d-9f73-41a4-9ba4-a443fcc0107d.png">

monicakogler:
we can add a little notif to the intercept header, and for speed let's do HTML confirm dialog ⚡ 

<img width="1334" alt="Screen Shot 2022-10-11 at 7 52 35 PM" src="https://user-images.githubusercontent.com/609038/195238738-6602e70f-921b-4af3-a48d-b97e39e216cc.png">

https://www.figma.com/file/9SS4DwqxNZT0vnRalOKa8e/mailing?node-id=1757%3A7911

psugihara:
released in 0.8.4 ✅

PR #260, opened 2022-10-17
Gh 245 safe delivery

psugihara:
## Change description

Adds the ability to force delivery of an intercepted send.

UI polish:
- global 1px translate for active state on buttons/links
- fixed height bug in PreviewViewer causing window to scroll
 
test at http://localhost:3883/intercepts/mock

Screen Shot 2022-10-16 at 9 37 19 PM
Screen Shot 2022-10-16 at 9 38 35 PM
Screen Shot 2022-10-16 at 9 39 00 PM

Screen Shot 2022-10-16 at 9 39 19 PM
Screen Shot 2022-10-16 at 9 39 48 PM

Screen Shot 2022-10-16 at 9 44 36 PM

## Issue link

#245 


alexfarrill:
Nice!

.eslintrc.json
```diff
@@ -54,6 +54,12 @@
         "@typescript-eslint/no-empty-function": [
           "error",
           { "allow": ["arrowFunctions"] }
+        ],
+        "react/no-unknown-property": [
```
alexfarrill, line 58: :tada

alexfarrill (changes requested):
When I click "force deliver" I see 404 and in server:
```
mailing 404 /api/intercepts/undefined/forceDeliver 116ms
```

alexfarrill (approved):
works now!  thanks

do we need a more contrasting button color for the "force send" button? (red?) It is kind of hidden

psugihara:
Yeah, good catch.

On design, I think it's correct for this to be subtle but legible (white on black is high contrast / legible). This is not something you will want to do very often but it will show every time you send and should not be a CTA. There are other designs in the Figma where it pops more, but I don't think that's the right hierarchy given the normal use-case for this page and dangerous consequences of clicking.  Monica thought a fade in would be nice and give a touch more prominence, I will try it.

alexfarrill:
Sounds good 👌

PR #266, opened 2022-10-18
Fix eslint warnings

elliothursh:
## Describe your changes
Fixes eslint warnings

## Issue link
N/A


packages/cli/src/commands/preview/server/livereload.ts
```diff
@@ -28,8 +28,17 @@ function requireParam(
...
+      throw new Error("No URL");
+    }
+    const url = new URL(req.url, `http://${req.headers.host}`);
+
+    const paramFromSearch = url.searchParams.get(param);
+    if (!paramFromSearch) {
+      throw new Error(`Missing required param: ${param}`);
+    }
+
+    return JSON.parse(paramFromSearch);
```
psugihara, line 41: Doesn't this do the same thing with more code? Maybe we should just disable that rule or I'm missing something? Doesn't really seem like it helps in the test case either.

elliothursh: 🤷 I dunno. In my mind it's kind of like casting something as any/unknown, seems better to avoid if possible?

.eslintrc.json
```diff
@@ -67,6 +67,16 @@
   "rules": {
     "semi": "off",
     "@next/next/google-font-display": "off",
+    "@next/next/no-html-link-for-pages": [
+      "error",
+      ["packages/cli/pages", "packages/web/pages"]
```
psugihara, line 72: cool

alexfarrill (approved):

packages/cli/src/commands/preview/server/livereload.ts
psugihara: oh ok, i had thought it was doing almost exactly the same thing with nicer syntax (asserting that it's not undefined)

psugihara (approved):

PR #267, opened 2022-10-18
disconnect prisma after tests

### LINKED ISSUE #261: integration tests will hit database connection limit with --watch
**Describe the bug**
ci tests will hit database connection limit with --watch.  It probably needs to close the database connection between runs or something

**To Reproduce**
- run yarn ci:test:integration:watch
- rerun tests a bunch of times (~10)
- you get a connection error

psugihara:
## Describe your changes

Good news: disconnecting prisma now and it doesn't hit the connection limit error
Bad news: after about 20 runs (better) it hits a heap error like this...

Screen Shot 2022-10-18 at 9 36 54 AM

I also
- switched the global name so we don't potentially overwrite the user's `global.prisma`.
- bumped jest versions (no dice on fixing the heap thing)

## Issue link

fixes #261 


alexfarrill (approved):
Great!  This error was annoying.  Too bad about the heap error.  Something else in the jest tests not getting cleaned up between runs maybe

PR #271, opened 2022-10-19
switch to importing assert rather than node:assert for node 14 compatibility

psugihara:
## Describe your changes

This fixes the build for node 14. I tested locally on 14, 16, 18 that this works.

`npx mailing deploy` is broken on 14 though because of a change to how npx handles `--yes`. I have a better idea for the deploy command, will fix that soon.

## Issue link

reported by lc on discord

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -110,7 +110,11 @@ export function buildSendMail(options: BuildSendMailOptions) {
     }
 
     if (forcePreview) {
-      log("💌 opening sendMail preview", mail);
+      const debugMail = {
+        ...mail,
+        html: "omitted",
+      };
+      log("💌 opening sendMail preview", debugMail);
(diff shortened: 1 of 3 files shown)


packages/core/src/index.ts
```diff
@@ -110,7 +110,11 @@ export function buildSendMail(options: BuildSendMailOptions) {
     }
 
     if (forcePreview) {
-      log("💌 opening sendMail preview", mail);
+      const debugMail = {
+        ...mail,
+        html: "omitted",
+      };
+      log("💌 opening sendMail preview", debugMail);
```
psugihara, line 117: snuck in this thing to make this log more readable

psugihara: i think it is much better because you can pull up the html in the previewer view anyway

elliothursh (approved):

PR #279, opened 2022-10-20
272 schema for list management

alexfarrill:
## Describe your changes
Before I go to much deeper, bunch of stuff you can do in here:

- POST to create a list
- POST to add a member to the list
- PATCH to change member's subscription status to the list (for unsub)
- GET your organization's lists
- GET members of a list

Also added some stuff I needed for:
- validation in api controllers

I think if you hook this up to sendMail as is and build an unsub tsx page you basically have an email list you can use with unsub, check it out.


packages/cli/prisma/schema.prisma
```diff
@@ -39,3 +40,21 @@ model ApiKey {
...
+  name           String
+  organizationId String
+  Organization   Organization @relation(fields: [organizationId], references: [id])
+  Member         Member[]
+
+  @@index([organizationId])
+}
+
+model Member {
+  id     String @id @default(cuid())
```
psugihara, line 55: @elliothursh i wonder if the Send model should reference this rather than "to" as a string

packages/cli/prisma/schema.prisma
```diff
@@ -39,3 +40,21 @@ model ApiKey {
...
+  email  String
+  listId String
+  List   List   @relation(fields: [listId], references: [id])
+  status String
```
psugihara, line 59: can we make status an enum?

packages/cli/src/pages/api/lists/[id]/members/[memberId].ts
```diff
@@ -0,0 +1,115 @@
...
+) {
+  const status = req.body.status;
+
+  const validatedMemberStatusInList = validateMemberStatusInList(status);
+
+  if (validatedMemberStatusInList.hasError) {
+    return validationErrorResponse(validatedMemberStatusInList, res);
+  }
+
+  // todo: why does this need to be updateMany?
```
Comment (psugihara) [inline packages/cli/src/pages/api/lists/[id]/members/[memberId].ts:30]: I think it's because there is not a unique constraint on listId + email

Schema needs `@@unique([listId, email])` for this to work

Comment (alexfarrill) [reply packages/cli/src/pages/api/lists/[id]/members/[memberId].ts]: ✔️ I think it's good to add that index so I added it, but you still have to use updateMany

Comment (psugihara) [reply packages/cli/src/pages/api/lists/[id]/members/[memberId].ts]: ftfy

packages/cli/src/pages/api/lists/[id]/members/[memberId].ts
```diff
@@ -0,0 +1,114 @@
...
+  const status = req.body.status;
+
+  const validatedMemberStatusInList = validateMemberStatusInList(status);
+
+  if (validatedMemberStatusInList.hasError) {
+    return validationErrorResponse(validatedMemberStatusInList, res);
+  }
+
+  await prisma.member.update({
+    where: { listId_email: { listId, email: memberId } },
```
Comment (psugihara) [inline packages/cli/src/pages/api/lists/[id]/members/[memberId].ts:31]: maybe missed forest for trees.

Is it sketchy to use email as the id here? It allows an attacker to unsub a bunch of people just by knowing their email. If we used a cuid it would remove that problem.

psugihara (approved):
Feels good. One thought about identifying list members (I think we should use the cuid vs email).

psugihara:
Elliot and I chatted briefly about how you would group related lists. I think you could just add a `listGroup String` field in the future to handle that.

packages/cli/src/pages/api/lists/[id]/members/[memberId].ts
Comment (alexfarrill) [reply packages/cli/src/pages/api/lists/[id]/members/[memberId].ts:31]: Yeah, true, I'll switch it to id.  The unsub link can have the id in the params or look it up from the send id

psugihara (approved):
looks good!

PR #280, opened 2022-10-21
support POST for /api/render

alexfarrill:
will fix this failing test after dinner and then it should be good


psugihara (approved):
awesome. nice tests.

alexfarrill:
@psugihara should be green now, I had to suck in some test util improvements I made in the other branch to get the GET to work and add a thing to generate the query string

PR #281, opened 2022-10-21
Single jest config

elliothursh:
## Describe your changes
I keep having an annoying time running tests while working in both next and non-next directories, because you have to run two separate commands depending on the directory a test is in.
This combines jest configs into a single config using 'projects', which allows both node and jsdom test to be run with yarn test.

## Issue link
N/A


psugihara (approved):
love it

alexfarrill:
Nice!

PR #283, opened 2022-10-21
Update integration test commands

elliothursh:
## Describe your changes
In trying to figure out how to run integration tests, I was a little confused by which integration test command i was supposed to be using.

This moves the "unsafe" commands into the test commands directly, and renames the tests you're supposed to run locally to test:integration and test:integration:watch.

It also adds the ability (in an unfortunately round about manner) to pass test flags to the jest command using `TEST_FLAGS="--watch apiKey" yarn test:integration`, this is nice imo if you're working on a specific test.

## Issue link
N/A


package.json
```diff
@@ -19,15 +19,13 @@
     "link:emails:js": "rm -rf emails; ln -s packages/cli/src/generator_templates/js/emails emails;",
     "build": "preconstruct build",
     "test": "yarn link:emails; yarn jest",
+    "test:integration": "./scripts/assert-free-ports.sh; MAILING_CI=true MAILING_DATABASE_URL=$MAILING_DATABASE_URL_TEST WEB_DATABASE_URL=$WEB_DATABASE_URL_TEST start-test ci:cli :3883 ci:web :3000 \"yarn link:emails; yarn jest -c jest.integration.config.ts $TEST_FLAGS\"",
```
alexfarrill, line 22: nice, I didn't know how to do this with the quotes, so that's why i moved it to the "unsafe" methods.  this is good that you can't run those commands without running the wrapping function that sets the env variables.  the point is to prevent you from running the truncate commands in the integration test setup on your dev database by accident.

package.json
```diff
@@ -19,15 +19,13 @@
...
+    "test:integration:watch": "./scripts/assert-free-ports.sh; MAILING_CI=true MAILING_DATABASE_URL=$MAILING_DATABASE_URL_TEST WEB_DATABASE_URL=$WEB_DATABASE_URL_TEST start-test ci:cli :3883 ci:web :3000 \"yarn link:emails; yarn jest -c jest.integration.config.ts --watch\"",
```
alexfarrill, line 23: @psugihara thought we needed --runInBand here for the tests to run serially (in addition to the related stuff in the jest config).  is that not so?

alexfarrill (approved):

package.json
elliothursh: setting maxWorkers to 1 in jest.integration.config.ts has the same effect as setting runInBand 👍

psugihara: Yeah, when we had tried before it didn't seem to apply correctly to `--watch`. But this seems to be working as expected for me 👍

Issue #284, opened 2022-10-22
Linter to ensure template components are named correctly


psugihara:
For collecting performance metrics about a template, we want to make sure templates and files are named the same thing.

Let's lint and throw friendly errors on 2 occasions:
- `npx mailing server build` should lint all templates
- when a preview is fetched in NODE_ENV=development

## Good ones

MyComponent.jsx
```tsx
export default function MyComponent() {
  return <Mjml></Mjml>
}
```

MyComponent.jsx
```tsx
const MyComponent = () => {
  return <Mjml></Mjml>
}
export default MyComponent;
```

## Bad ones

MyComponent.jsx
```tsx
export default function () {
  return <Mjml></Mjml>
}
```

MyComponent.jsx
```tsx
export default function NotMyComponent() {
  return <Mjml></Mjml>
}
```

MyComponent.jsx
```tsx
const NotMyComponent = () => {
  return <Mjml></Mjml>
}
export default NotMyComponent;
```

PR #287, opened 2022-10-25
284 linter to ensure template components are named correctly

psugihara:
## Describe your changes

Adds a linter for checking things about the templates / manifest. The only check here is that the filename matches the function name.

This helps ensure #248 will work as expected because that's the only place where we derive template name from function name rather than filename.

https://github.com/sofn-xyz/mailing/pull/248/files#r1004620077

Screen Shot 2022-10-24 at 4 11 36 PM
Screen Shot 2022-10-24 at 4 07 47 PM

## Issue link

#284 


psugihara:
re-opening as I can't figure out a less kludgy way to protect from this mismatch https://github.com/sofn-xyz/mailing/pull/248/files#r1004620077

packages/cli/tsconfig.json
```diff
@@ -18,7 +18,10 @@
     "noEmit": true,
     "incremental": true,
     "resolveJsonModule": true,
-    "baseUrl": "./"
+    "baseUrl": "./",
+    "paths": {
```
elliothursh, line 22: did you mean to include this?

elliothursh (approved):

packages/cli/tsconfig.json
psugihara: nah, reverting that one. thx

PR #292, opened 2022-10-27
Bugfix assert free ports

alexfarrill:
## Describe your changes

In main, `./scripts/assert-free-ports.sh` does not work as intended:  
- If you have web or cli running it prints "Refusing to run because port 3883 is already in use" but exits with exit code 0 (success).  It should do the opposite (error code 0 if no servers are running, 1 if servers running)
- the commands following it should be correctly checking its return type by using `&&` not `;`.  In main the yarn commands that use the script print an error message but continue on to run when there is a server running.

In the branch the behavior changes to:

If you have web or cli running you should:
```
./scripts/assert-free-ports.sh; echo $?
Refusing to run because port 3883 is already in use
1
```
and commands like yarn test:integration should fail early

Otherwise you should see:
```
./scripts/assert-free-ports.sh; echo $?
0
```
and commands like yarn test:integration should run


scripts/assert-free-ports.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 
-lsof -i :3000 > /dev/null && echo 'Refusing to run because port 3000 is already in use' && exit 0
-lsof -i :3883 > /dev/null && echo 'Refusing to run because port 3883 is already in use' && exit 0
\ No newline at end of file
+lsof -i :3000 > /dev/null && echo 'Refusing to run because port 3000 is already in use' && exit 1
+lsof -i :3883 > /dev/null && echo 'Refusing to run because port 3883 is already in use' && exit 1
+
+exit 0
```
alexfarrill, line 6: without this, the exit code is 1 when neither lines 3 nor 4 obtain.  I am not great at bash so I couldn't tell you why, but you can verify by running:
`./scripts/assert-free-ports.sh; echo $?`

If you have web or cli running you should see:
```
Refusing to run because port 3883 is already in use
1
```

Otherwise you should see:
```
./scripts/assert-free-ports.sh; echo $?
0
```

package.json
```diff
@@ -19,13 +19,14 @@
...
-    "ci:test:integration": "./scripts/assert-free-ports.sh; MAILING_CI=true MAILING_DATABASE_URL=$MAILING_DATABASE_URL_TEST WEB_DATABASE_URL=$WEB_DATABASE_URL_TEST start-test ci:cli :3883 ci:web :3000 \"yarn link:emails; yarn jest -c jest.integration.config.ts\"",
+    "test:integration": "./scripts/assert-free-ports.sh && yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts $TEST_FLAGS\"",
+    "test:integration:watch": "./scripts/assert-free-ports.sh && yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts --watch\"",
+    "ci:test": "yarn integration:servers:start test",
+    "ci:test:integration": "./scripts/assert-free-ports.sh && yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts\"",
     "ci:cli": "cd packages/cli && yarn ci:server",
     "ci:web": "cd packages/web && yarn ci:server",
     "test:e2e": "bundle && bundle exec ruby e2e/cli.rb run-all",
+    "integration:servers:start": "MAILING_CI=true MAILING_DATABASE_URL=$MAILING_DATABASE_URL_TEST WEB_DATABASE_URL=$WEB_DATABASE_URL_TEST start-test ci:cli :3883 ci:web :3000",
```
alexfarrill, line 29: @elliothursh give me your opinion naming this, but this is useful to have

package.json
```diff
@@ -19,13 +19,14 @@
...
-    "test:integration:watch": "./scripts/assert-free-ports.sh; MAILING_CI=true MAILING_DATABASE_URL=$MAILING_DATABASE_URL_TEST WEB_DATABASE_URL=$WEB_DATABASE_URL_TEST start-test ci:cli :3883 ci:web :3000 \"yarn link:emails; yarn jest -c jest.integration.config.ts --watch\"",
-    "ci:test": "MAILING_DATABASE_URL=$MAILING_DATABASE_URL_TEST WEB_DATABASE_URL=$WEB_DATABASE_URL_TEST start-test ci:cli :3883 ci:web :3000 test",
-    "ci:test:integration": "./scripts/assert-free-ports.sh; MAILING_CI=true MAILING_DATABASE_URL=$MAILING_DATABASE_URL_TEST WEB_DATABASE_URL=$WEB_DATABASE_URL_TEST start-test ci:cli :3883 ci:web :3000 \"yarn link:emails; yarn jest -c jest.integration.config.ts\"",
+    "test:integration": "./scripts/assert-free-ports.sh && yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts $TEST_FLAGS\"",
+    "test:integration:watch": "./scripts/assert-free-ports.sh && yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts --watch\"",
+    "ci:test": "yarn integration:servers:start test",
```
alexfarrill, line 24: the other 3 commands are no-ops but this one technically now has MAILING_CI=true and did not before.  We may have omitted that by mistake?

psugihara (approved):
nice

PR #295, opened 2022-10-27
257 add cypress tests that use the integration db

alexfarrill:
## Describe your changes
Add cypress test for login and fix /account page which is broken in main


.github/workflows/cypress_integration.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Use Node.js 18.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: "yarn"
+      - name: Install dependencies
+        run: yarn
+      - name: Build the app
+        run: yarn build
+      - name: Cypress run integration tests
```
alexfarrill, line 43: @elliothursh how do you do this :)  

also, do you think the jest integration and cypress integration be combined into one action to avoid multiple postgres setup?  i think this is better (separate) so they're not executing serially / dependent on each other, but what do you think

.github/workflows/cypress_integration.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: cypress-io/github-action@v4
+        run: test:integration:cypress
+        # env:
+        #   DEBUG: "@cypress/github-action"
+        # with:
+        #   install: false
+        #   start: yarn ci:server:nohup
+        #   working-directory: packages/cli
```
alexfarrill, line 51: if the config is supposed to look more like this (from the regular cypress config), then I think we just have to specify a different start command and the other cypress config

packages/cli/src/pages/settings.tsx
```diff
@@ -18,6 +18,7 @@ export const getServerSideProps = withSessionSsr<{ user: any }>(
 
     const apiKeys: ApiKey[] = await prisma.apiKey.findMany({
       where: { organizationId: user.organizationId },
+      select: { id: true, active: true },
```
alexfarrill, line 21: fixes a bug in main where you can't login after signing up

packages/web/package.json
```diff
@@ -33,7 +33,6 @@
     "jest-mock-extended": "^2.0.7",
     "postcss": "^8.4.14",
     "prisma": "^4.2.0",
-    "tailwindcss": "^3.1.6",
-    "typescript": "^4.7.4"
```
alexfarrill, line 37: typescript is also declared in dependencies, so this is a dupe that emits a warning in main

packages/cli/cypressIntegration.config.ts
```diff
@@ -0,0 +1,48 @@
...
+    baseUrl: "http://localhost:3883",
+    setupNodeEvents(on, _config) {
+      on("task", {
+        "db:reset": resetDb,
+      });
+    },
+  },
+});
+
+// copied from testUtilIntegration.ts
```
alexfarrill, line 24: for some reason you cannot import these modules from "../../testUtilIntegration", it will cause cypress to throw an error that the config file is invalid.  i think because ../.. is outside of packages/cli 🤷

alexfarrill:
```$ /Users/alexfarrill/projects/mailing/node_modules/.bin/cypress run -C cypressIntegration.config.ts
Your configFile is invalid: /Users/alexfarrill/projects/mailing/packages/cli/cypressIntegration.config.ts

It threw an error when required, check the stack trace below:

Error: Cannot find module 'file:///Users/alexfarrill/projects/mailing/packages/cli/cypressIntegration.config.ts'
Require stack:
- /Users/alexfarrill/Library/Caches/Cypress/10.9.0/Cypress.app/Contents/Resources/app/packages/server/lib/plugins/child/run_require_async_child.js
- /Users/alexfarrill/Library/Caches/Cypress/10.9.0/Cypress.app/Contents/Resources/app/packages/server/lib/plugins/child/require_async_child.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:985:15)
    at Function.Module._resolveFilename (/Users/alexfarrill/Library/Caches/Cypress/10.9.0/Cypress.app/Contents/Resources/app/node_modules/tsconfig-paths/lib/register.js:75:40)
    at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename] (/Users/alexfarrill/Library/Caches/Cypress/10.9.0/Cypress.app/Contents/Resources/app/node_modules/@cspotcode/source-map-support/source-map-support.js:811:30)
    at Function.Module._load (node:internal/modules/cjs/loader:833:27)
    at Module.require (node:internal/modules/cjs/loader:1057:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at /Users/alexfarrill/Library/Caches/Cypress/10.9.0/Cypress.app/Contents/Resources/app/packages/server/lib/plugins/child/run_require_async_child.js:106:14
    at processTicksAndRejections (node:internal/process/task_queues:96:5)```

alexfarrill: however it is fine to import a type from there (makes sense since only tsc is involved and cypress can ignore)
also ok to import `import webPrisma from "../../packages/web/prisma";`  (maybe because it's a sibling directory, not a parent?)

.github/workflows/cypress_integration.yml
elliothursh: This seems good to me 👍

.github/workflows/cypress_integration.yml
elliothursh: Seems like it's working without the cypress action, seems chill 👍

elliothursh (approved):

PR #305, opened 2022-11-02
Navigation

psugihara:
## Describe your changes

This adds the new navigation for Mailing Platform. It's behind a feature flag for now so that we can merge it in because it requires some layout changes that will get merge conflicty.

The idea is that it will show in production deployed mode and hide in dev. If there's not a db hooked up, we'd want to point you to docs / tell you about the Audiences/Unsub/Analytics features you'll get by hooking up a db.

You can add nav=1 to the url querystring to see it on the left side.

Screenshot 2022-11-02 at 3 23 30 PM


packages/cli/src/components/IndexPane/hooks/usePreviewTree.ts
```diff
@@ -90,14 +90,18 @@ export function usePreviewTree(
...
 
   useEffect(() => {
-    if (cursor === -1 || router?.asPath === routes[cursor]?.path) return;
-    safeReplaceState(router, routes[cursor]?.path);
+    const [path, qs] = router.asPath.split("?");
+    if (cursor === -1 || path === routes[cursor]?.path) return;
+    safeReplaceState(
+      router,
+      qs ? `${routes[cursor]?.path}?${qs}` : routes[cursor]?.path
+    );
```
psugihara, line 104: little fix, this file was not handling querystrings correctly

alexfarrill (approved):

PR #306, opened 2022-11-03
3x faster parsing of html -> text

psugihara:
## Describe your changes

Shower idea just now. This switches the library used to parse html and get the text preview for the client view. We end up hitting this api a lot and there's no pagination.

With the 25 Fynn previews, this takes latency from 1-3s to <300ms. Cool!
Props to https://github.com/taoqf/node-html-parser @taoqf

Screen Shot 2022-11-02 at 8 05 18 PM

investigating e2e tests


alexfarrill (approved):
Nice one!

Issue #150, opened 2022-08-27
MDX support


agamm:
It would be nice to be able to transform MDX files into compatible emails and get all the benefits of mailing.

My current use case is to use mdx for a blog and a newsletter where I would like to write in mdx and then send it out preferably via mailing.

I can imagine I'd like to use mjml as a base template where I insert the mdx content in between.

mrlubos:
Hi @agamm! Could you please describe in more detail how do you imagine this feature working? Is it currently throwing any errors when you try to implement the MDX library? Wondering if this is related to #93 and #129 at all. Thanks!

psugihara:
I've heard this request from a few more users now and after using mdx a bit have a better idea of how this could work. Would love some feedback on whether this design would be useful to people:

- Add `templateName?: string` and `props?: any` to `sendMail`'s [`ComponentMail` arg type](https://github.com/sofn-xyz/mailing/blob/main/packages/core/src/index.ts#L10)

```ts
await sendMail({
  templateName: "MyMDXTemplate", // refers to emails/MyMDXTemplate.mdx
  props: { name: "hey I'm an example prop" },
  to: "example.user@gmail.com"
})
```

With this API design we make a trade-off between strong prop types and ease of installation because I don't want the user to have to install an MDX webpack loader and get the MDX components to import correctly in their app code (I think they can already do that if they want and pass the component into `sendMail` as `component` as long as they use MJML components).

- Add support for different "loaders" that would render templates differently based on file extension of the template in your `emailsDir`. We should build in such a way that we can add more loaders easily.
- Our current mjml-react based loader would run on .tsx and .jsx extensions
- A new mdx loader would run on .mdx files
  - The mdx should compile to mjml-react then html via an [mdx provider](https://mdxjs.com/docs/using-mdx/#mdx-provider) or the [components prop](https://mdxjs.com/docs/using-mdx/#components). That way we can provide email friendly stock implementations of all of the markdown components [here](https://mdxjs.com/table-of-components/).
  - If the user passes `components` in the prop to sendmail, we should merge them with our system default `mjml-react` components (allow the user implementation to override the mailing defaults).
- Include the mdx loader by default so that you don't have to install anything else (re-evaluate this if it adds unreasonable weight to the bundle, but I think it's probably fine)

PR #309, opened 2022-11-03
swc/register to esbuild

psugihara:
## Describe your changes

I was investigating markdown support and one blocker was that there's no mdx support for swc. However, [mdx does support esbuild](https://mdxjs.com/docs/getting-started/#esbuild) which we were using for bundling already. This swaps the register calls to use esbuild rather than swc... these hooks are used when ts is required at runtime.

## Issue link

#150 


alexfarrill (approved):
great, simplifies things and esbuild is very fast

PR #310, opened 2022-11-04
email preferences

alexfarrill:


packages/cli/src/generator_templates/js/emails/components/Footer.jsx
```diff
@@ -12,7 +12,7 @@ export default function Footer() {
           color={grayDark}
         >
           © {new Date().getFullYear()} BookBook&nbsp;&nbsp;·&nbsp;&nbsp;
-          <a href="https://www.mailing.run" target="_blank" rel="noreferrer">
+          <a href="MM_EMAIL_PREFERENCES_URL" target="_blank" rel="noreferrer">
```
alexfarrill, line 15: It would be cool if in development mode, this would do something special when you click like open a page that explains how it will work for you in production

packages/core/src/index.ts
```diff
@@ -158,20 +160,30 @@ export function buildSendMail<T>(options: BuildSendMailOptions<T>) {
...
+
           const stringHtml = mailOptions.html?.toString();
           if (stringHtml) {
             mailOptions.html = instrumentHtml({
               html: stringHtml,
               messageId: messageId,
               apiUrl: MAILING_API_KEY,
             });
+
+            stringHtml.replace(
```
alexfarrill, line 181: @psugihara what's the best way to verify this works manually?

packages/cli/src/pages/api/messages/index.ts
```diff
@@ -15,7 +17,94 @@ export default async function handler(
     return res.status(405).json({ error: "Method not allowed" });
   }
 
-  if (!(await validateApiKey(req, res))) return;
+  const apiKey = apiKeyFromReq(req);
```
alexfarrill, line 20: should write some tests for this file.  this logic look right though?

alexfarrill:
- [ ] fix tests
- [ ] sendMail can optionally take a list and it does what it's supposed to

packages/cli/src/pages/api/messages/index.ts
```diff
@@ -15,7 +17,94 @@ export default async function handler(
...
+    });
+  }
+
+  // return early if your status to either list is not "subscribed"
+  if (
+    listMember?.status !== "subscribed" ||
+    defaultListMember?.status !== "subscribed"
+  ) {
+    res.status(200).json({ message: "user is not subscribed to either list" });
+  }
```
psugihara, line 81: - needs a different condition, should be false if user is not a member of the list
- should return
- message is supposed to be a model, use error

```suggestion
  if (
    listMember?.status === "unsubscribed" ||
    defaultListMember?.status === "unsubscribed"
  ) {
    res.status(200).json({ error: "user is not subscribed to either list" });
    return;
  }
```

packages/cli/src/pages/api/messages/index.ts
psugihara: yeah looks right. one little error I found.

packages/core/src/index.ts
psugihara: i think this could work:
- add a template with an unsubscribe link
- use the PreviewSender to manually send yourself one

psugihara:
idk why it printed my review twice

packages/cli/src/pages/api/lists/[id]/subscribe.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (email) {
+      try {
+        await prisma.member.create({
+          data: { listId, email, status: "subscribed" },
+        });
+
+        return res.status(201).end();
+      } catch (error) {
+        if (error instanceof Prisma.PrismaClientKnownRequestError) {
+          if (error.code === "P2002") {
```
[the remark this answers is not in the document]
Comment (alexfarrill) [reply packages/cli/src/pages/api/lists/[id]/subscribe.ts:31]: Will you also try to trigger this? (open up /subscribe/listId and submit the same email twice and it should hit this).  When I last tried it, `error instanceof Prisma.PrismaClientKnownRequestError` always returned false

alexfarrill:
@psugihara if you have a sec, can you figure out why these files are not excluded from being copied into .mailing?

```
alexfarrill:mailing(289-unsub-page) find ./packages/cli/.mailing -name __integration__
./packages/cli/.mailing/src/pages/api/messages/__integration__
./packages/cli/.mailing/src/pages/api/lists/__integration__
./packages/cli/.mailing/src/pages/api/lists/[id]/members/__integration__
./packages/cli/.mailing/src/pages/api/lists/[id]/__integration__
./packages/cli/.mailing/src/pages/api/apiKeys/__integration__
./packages/cli/.mailing/src/pages/api/hooks/__integration__
./packages/cli/.mailing/src/pages/api/__integration__
```

We already have __integration__ in the regex exclude thing, yet somehow they are getting in there.  resolving will fix the jest tests

alexfarrill:
Waah, in setup.ts, in the copy block for MM_DEV on line :161, those paths are only the top level directories, that matcher does not filter every file.  I am not sure the behavior in the else block

psugihara:
Ah dang. Well good find. I'm guessing it hasn't really caused any problems.

alexfarrill:
- [ ] put behind a feature flag
- [ ] Setup a mailing deployment inside of web
- [ ] Write API docs in a github readme
- [ ] Write mini doc on lists and the unsubscribe feature
- [ ] bulk add mailing mailing list users

alexfarrill:
@psugihara @elliothursh lmk what you think, especially about naming of /unsubscribe/[memberId] and the api endpoints

alexfarrill:
- [ ] @psugihara @monicakogler template and draft for first email to mailing

packages/cli/src/pages/audiences/index.tsx
```diff
@@ -132,6 +135,7 @@ const PreviewIndex: NextPage<AudiencesProps> = ({
   sortOrder,
   page,
 }) => {
+  console.log("page", page);
```
alexfarrill, line 138: rm

packages/cli/src/pages/audiences/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+          destination: "/audiences",
+          permanent: false,
+        },
+      };
+    }
+
+    let defaultListId: string | undefined = members[0]?.listId;
+    if (!defaultListId) {
+      // no members, so find the default list
+      defaultListId = (
```
psugihara, line 77: refactor idea...

defaultListId = (await findOrCreateDefaultList).id

I think we do a similar thing in a bunch of places.

packages/core/src/util/log.ts
```diff
@@ -9,6 +9,10 @@ export function log(message?: any, ...optionalParams: any[]) {
 }
 
 export function error(message?: any, ...optionalParams: any[]) {
+  // e.g. when there is a jest test like instrumentHtml.test.ts
+  // that intentionally throws an error, don't log from a
+  // try-catch that logs and re-raises the error
+  if ("test" === NODE_ENV) return;
```
psugihara, line 15: another way to achieve is a spy on console.error. If we wanted this globally, could do it in the setup.

alexfarrill:
- [ ] the unsubscribe page should not show example@example.com

psugihara:
A couple more things working through this in more detail with the doc...

- [ ] currently the sendMail calls needs a listId so they can't actually create lists dynamically and the callsites are hard to understand because you'll just have CUIDs in the code. I think sendMail should take a unique list name instead and create it if it doesn't exist (case 1 step A in the unsub link API [here](https://docs.google.com/document/d/1sZexbNUsgqIddegpc6-aN4Dp8N6woM75H7ZixO9TZTQ/edit?usp=sharing)).
- [ ] test to ensure 4.a. works? maybe I'm missing it

Screenshot 2022-11-09 at 5 05 46 PM

alexfarrill:
- [ ] create list from sendMail
- [ ] edit page for list so you can edit the displayName

alexfarrill:
- [ ] remove feature flag before merging

Issue #289, opened 2022-10-25
unsub page


alexfarrill:
A page where you can unsub from lists, we will later include this in a magic link that you can import to your email templates

alexfarrill:
## unsub tsx
url: /unsubscribe/[memberId]
- [ ] get the listMember or 404 if it's not found
- [ ] render the list of lists that can be unsubscribed
- [ ] get the email's status for each list
- [ ] always render an unsubscribe all
- [ ] do the right thing when the form is submitted
- [ ] show a thank you message when the form is submitted
- [ ] add a noindex header to this page

## magic link
tbd...

PR #311, opened 2022-11-05
log in / sign up styling

monicakogler:
## Describe your changes
makes input and button components + general form styling

## TODO

- [x] load logo
- [x] make refs work in input component


psugihara:
going to fix this quick so we can use in #289

PR #313, opened 2022-11-07
Fix next 13 errors

psugihara:
## Describe your changes

Fixes a few errors when running mailing with next13.

Not sure why the e2e CI doesn't catch these.


alexfarrill (approved):
hmm, i'm not sure why either.  does server build fail on next13 or do the links just not work?

PR #321, opened 2022-11-08
Intercept force deliver bugfix

### LINKED ISSUE #316: mailing caught error Error: getaddrinfo ENOTFOUND smtp.example.com
**Describe the bug**
Not able to get the test server to force send with the correct config settings, maybe im missing something. 
I tried changing /emails/index.ts transport, but still getting the following error.

```sh
mailing caught error Error: getaddrinfo ENOTFOUND smtp.example.com
```

even with /emails/index.ts config
```ts
const transport = nodemailer.createTransport({
  host: 'smtp.mailtrap.io',
  port: 2525,
  auth: {
    user: 'username',
    pass: 'password'
  }
});
```

**To Reproduce**

I must be missing a config step, but i updated the transport config in the /emails/index.ts, and when I run `npx mailing`, then send 

1. run npx mailing to init project inside nextjs app
2. change the config in emails/index.ts in transport ensuring host is different
3. restart mailing server
4. trigger send email within main nextjs app so it gets intercepted by mailing server
5. Click on 'Force send'
6. See error
```
mailing caught error Error: getaddrinfo ENOTFOUND smtp.example.com
```

**Expected behavior**
To send email to mailtrap

I was able to get it to send within the main nextjs app by doing  
dangerouslyForceDeliver: true 
inside sendMail

Any help would be appreciated

psugihara:
## Describe your changes

This fixes a bug with the Force Deliver Intercepts feature. It was using the example copy of sendMail bundled by esbuild rather than the user-defined sendMail. I fixed by getting rid of the intercept deliver route and instead using /api/sendMail. I think this is a better design because it gets rid of a codepath.

I added a unit test for the auth change (api key checks always pass in development so that the localhost preview server can use any api without restriction).

I manually tested with an example next app pointed at mailtrap and it seems to work well.

Screenshot 2022-11-08 at 1 10 36 PM

Screenshot 2022-11-08 at 12 55 38 PM

Screenshot 2022-11-08 at 12 53 24 PM

## Issue link

fixes #316 

### DIFF
--- a/.changeset/grumpy-crabs-itch.md
+++ b/.changeset/grumpy-crabs-itch.md
@@ -1,0 +1,7 @@
+---
+"mailing": patch
+"mailing-core": patch
+"web": patch
+---
+
+bugfix for force delivery from intercepts #316
--- a/.changeset/slimy-students-melt.md
+++ b/.changeset/slimy-students-melt.md
@@ -1,7 +1,0 @@
----
-"mailing": patch
-"web": patch
-"mailing-core": patch
----
-
-fix next 13 errors
--- a/.changeset/stale-teachers-begin.md
+++ b/.changeset/stale-teachers-begin.md
@@ -1,6 +1,0 @@
----
-"mailing": patch
-"mailing-core": patch
----
-
-remove unused images, fix test file ignores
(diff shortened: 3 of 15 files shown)


alexfarrill (approved):
🌶️

PR #323, opened 2022-11-09
Make mobile previewer iphone 8 sized

elliothursh:
## Describe your changes
Increases the viewport size of the mobile preview to 375x667, which is more inline with the larger viewport sizes of recent phones. 

## Issue link
N/A


psugihara (approved):
nice

alexfarrill (approved):

Issue #322, opened 2022-11-09
422 on force send


SSardorf:
**Describe the bug**
I am getting a 422 error in the terminal when I force send.

I am also getting the following error popup:
<img width="441" alt="image" src="https://user-images.githubusercontent.com/22572423/200796345-e02f7ae9-58b7-4c69-a7ff-44ec71b6df68.png">

This is my API route that triggers the sending of the email
```
//api/sendTestEmail.tsx
import sendMail from "emails";
import ConfirmEmail from "emails/ConfirmEmail";

export default async function handler(req: any, res: any) {

  sendMail({
      subject: "First Test Email",
      to: "personalEmail@hidden.com",
      component: <ConfirmEmail firstName="Amelita" ctaText={"Verify this test email"} callbackUrl={"https://callbackUrl.com/"} />
    });
  }
  ```

This is the my emails/index.ts
```
import nodemailer from "nodemailer";
import { buildSendMail } from "mailing-core";

const transport = nodemailer.createTransport({
  pool: true,
  host: "smtp.postmarkapp.com",
  port: 587,
  secure: true, // use TLS
  auth: {
    user: process.env.POSTMARK_CLIENT,
    pass: process.env.POSTMARK_CLIENT,
  },
});

const sendMail = buildSendMail({
  transport,
  defaultFrom: "workEmail@hidden.com",
  configPath: "./mailing.config.json",
});

export default sendMail;

```

Finally, this is the preview output
```
mailing 💌 opening sendMail preview {
  subject: 'First Test Email',
  to: 'personalEmail@hidden.com',
  component: {
    '$$typeof': Symbol(react.element),
    type: [Function: ConfirmEmail],
    key: null,
    ref: null,
    props: {
      firstName: 'Amelita',
      ctaText: 'Verify this test email',
      callbackUrl: 'https://callbackUrl.com/'
    },
    _owner: null,
    _store: {}
  },
  html: 'omitted'
}
API resolved without sending a response for /api/sendTestEmail, this may result in stalled requests.
mailing 💌 opening sendMail preview {
  subject: 'First Test Email',
  to: 'personalEmail@hidden.com',
  component: {
    '$$typeof': Symbol(react.element),
    type: [Function: ConfirmEmail],
    key: null,
    ref: null,
    props: {
      firstName: 'Amelita',
      ctaText: 'Verify this test email',
      callbackUrl: 'https://callbackUrl.com/'
    },
    _owner: null,
    _store: {}
  },
[...]

PR #325, opened 2022-11-09
Gh 322 422 on force send

psugihara:
## Describe your changes

Turns out NODE_ENV is just not set in a lot of environments. My bad. This should be a more universal fix.

Published to npm as 0.8.16-next.0 and this fix works in the context I repro'd the failure (NODE_ENV not set 🤦).

## Issue link

#322 

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -97,8 +97,8 @@
   },
   "scripts": {
     "dev": "rm -rf .mailing && yarn prisma migrate dev && src/dev.js",
-    "ci:server": "rm -rf .mailing && yarn prisma generate && yarn prisma migrate deploy && NODE_ENV=test src/dev.js > test.log",
-    "ci:server:nohup": "rm -rf .mailing && NODE_ENV=test nohup src/dev.js --quiet > /dev/null 2>&1 &"
+    "ci:server": "rm -rf .mailing && yarn prisma generate && yarn prisma migrate deploy && NEXT_PUBLIC_MAILING_SKIP_AUTH=false NODE_ENV=test src/dev.js > test.log",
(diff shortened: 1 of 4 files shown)


packages/cli/package.json
```diff
@@ -97,8 +97,8 @@
   },
   "scripts": {
     "dev": "rm -rf .mailing && yarn prisma migrate dev && src/dev.js",
-    "ci:server": "rm -rf .mailing && yarn prisma generate && yarn prisma migrate deploy && NODE_ENV=test src/dev.js > test.log",
-    "ci:server:nohup": "rm -rf .mailing && NODE_ENV=test nohup src/dev.js --quiet > /dev/null 2>&1 &"
+    "ci:server": "rm -rf .mailing && yarn prisma generate && yarn prisma migrate deploy && NEXT_PUBLIC_MAILING_SKIP_AUTH=false NODE_ENV=test src/dev.js > test.log",
+    "ci:server:nohup": "rm -rf .mailing && NEXT_PUBLIC_MAILING_SKIP_AUTH=false NODE_ENV=test nohup src/dev.js --quiet > /dev/null 2>&1 &"
```
psugihara, line 101: this makes it so that in integration tests, api key is required

elie222 (approved):

packages/cli/package.json
alexfarrill: Does `yarn test:integration` work locally under this scheme?  You may also have to set this variable in `integration:servers:start`

psugihara: yep, will update

psugihara: ah nvm, that just calls this

alexfarrill (approved):

PR #328, opened 2022-11-10
289 unsub page

alexfarrill:


alexfarrill:
## tests
- [ ] navbar should not show on the unsubscribe page
- [ ] sendMail works with a list and obeys rules about unsubscribe link being included
- [ ] sendMail case 2 it should throw an error when trying to sendMail to a listId without an unsubscribe link in the template
- [ ] sendMail case 4 Call without list and template has no magic unsub link -- Send it! Users should do this for emails that cannot be unsubscribed from (e.g. order confirmation, password reset)
- [ ] sendMail creates a new (non-default) list if listName doesn't exist in the db yet

## update docs
- [ ] you can pass a listName and it will create everything for you
- [ ] you can't create lists on the settings page anymore

alexfarrill (commented):
@psugihara take a look at this changeset https://github.com/sofn-xyz/mailing/pull/328/commits/2a864367d9e13e57f9381d1dd4f09013025731eb

although a bit hacky, it should be safe (we check process.env.MAILING_CI and a magic string for api key).  This is necessary because buildSendMail checks process.env.MAILING_API_KEY so that has to be set before the integration test server is started.  It won't work to make an API key after the server is started.  However, this lets us test a lot of the logic we added to sendMail and there are some bugs fixed in here that I found through the tests.  Still have a few more tests to add.

psugihara:
Looks good, I think the magic key is a fine solution. We should leave some tests for the case where that is not set though. Could either set MAILING_API_KEY and MAILING_API_URL in the test (e.g. `(process.env as any).MAILING_API_KEY = "testApiKey";`) or delete it temporarily in the test to make sure the basic send codepath works.

alexfarrill:
Yup, i had the same thought, the tests in this block should cover that: https://github.com/sofn-xyz/mailing/pull/328/files#diff-70dbc1c6d43da17fb584942d3f12a184fbc2217dc99a84f3a897cebda8b822efR22

we can also delete the process variables in that block

PR #329, opened 2022-11-10
remove logger use on client

psugihara:
## Describe your changes

I think this will error when it's called because the logger requires fs. This is the only instance I can find where it's used on the client.


alexfarrill (approved):
https://www.youtube.com/watch?v=AjPau5QYtYs 👍

Issue #331, opened 2022-11-11
How can I query a headless CMS to fetch data to be passed in Previews ?


jbtheard:
**Is your feature request related to a problem? Please describe.**
My emails content is currently stored in a headless CMS so that marketer can easily edit them.

**Describe the solution you'd like**
A way to retrieve content via GraphQL so that I can build previews with data from the CMS instead of mocking the data.

**Describe alternatives you've considered**
Manually copy/pasting result of query as  source data

**Additional context**
Perhaps we could query data from each Preview file ? [Next 13](https://beta.nextjs.org/docs/rendering/server-and-client-components) may allow that ?

psugihara:
I like this idea, @jbtheard! I've heard similar requirements a couple times now so I'd like to prioritize a solution. I think we should support async Preview functions so that you could prepare whatever data you need.

That way you could do something like this....


```tsx
export async function welcome() {
  const res = await fetch("my.cool.api/data.json");
  const data = await res.json()
  return <Welcome data={data} />
}
```

Would that work for you?

We do a 1-pass render via mjml-react so hooks or more familiar react solutions are hard to implement, but this would be relatively easy.

jbtheard:
You're absolutely right; using a hook would feel more familiar but your suggestion would absolutely work and fit the need. I'm happy to test it if/when sth becomes available on the matter.
So happy this could land on the backlog soon!

psugihara:
released in v0.8.18 

LMK if you have any issues @jbtheard. Thanks for your help understanding the use-case!

jbtheard:
I confirm that's working brilliantly! Thanks a lot, really helpful.
Now very close to being able to create emails on my headless CMS and automatically update template in my CRM

psugihara:
Very cool. Would love to read a mini write-up about your setup if you ever have the chance to document how you're doing it!

jbtheard:
Will definitely try to take time to do that! (likely next week though)

Issue #276, opened 2022-10-20
Cannot call `useEffect` in preview component


klapec:
**Describe the bug**
I wish to be able to call `useEffect` in my code, particularly in the preview `./emails/previews/Foo.tsx` component. I'm not able though - I'm getting this error:
```
@org/mailing:dev: error - ../../../node_modules/react/cjs/react.development.js (1634:0) @ useEffect
@org/mailing:dev: error - TypeError: Cannot read properties of null (reading 'useEffect')
@org/mailing:dev:     at useEffect (webpack-internal:///../../../node_modules/react/cjs/react.development.js:1635:21)
@org/mailing:dev:     at Object.Preview (webpack-internal:///./src/moduleManifest.js:663:76)
@org/mailing:dev:     at getPreviewComponent (webpack-internal:///./src/util/moduleManifestUtil.ts:24:43)
@org/mailing:dev:     at getStaticProps (webpack-internal:///./src/pages/previews/[[...path]].tsx:53:104)
@org/mailing:dev:     at Object.renderToHTML (/Volumes/Dev/repos/org-frontend-monorepo/node_modules/next/server/render.tsx:749:20)
@org/mailing:dev:     at runMicrotasks (<anonymous>)
@org/mailing:dev:     at processTicksAndRejections (node:internal/process/task_queues:96:5)
@org/mailing:dev:     at doRender (/Volumes/Dev/repos/org-frontend-monorepo/node_modules/next/server/base-server.ts:1288:30)
@org/mailing:dev:     at cacheEntry.responseCache.get.isManualRevalidate.isManualRevalidate (/Volumes/Dev/repos/org-frontend-monorepo/node_modules/next/server/base-server.ts:1429:24)
@org/mailing:dev:     at /Volumes/Dev/repos/org-frontend-monorepo/node_modules/next/server/response-cache/index.ts:117:28 {
@org/mailing:dev:   page: '/previews/[[...path]]'
@org/mailing:dev: }
```

**To Reproduce**

1. Create a preview component for a template
2. Add `import { useEffect } from 'react';`
3. Call
```
useEffect(() => {
    console.log(1);
  }, []);
```
within preview component's body
4. Crash

**Expected behavior**
Calling hooks should be allowed.

**Desktop (please complete the following information):**

- OS: [e.g. iOS] macos@12.5.1
- Browser [e.g. chrome, safari] irrelevant but Chrome@106.0.5249.119
- Version [e.g. 22] 0.8.7

[...]

PR #336, opened 2022-11-14
support for async preview functions

psugihara:
## Describe your changes

I think it's just this easy. This adds support for making preview functions async so that you can do data fetching inside of them. This is completely optional functionality so I'm going to push off documenting for the forthcoming docs site.

I added a single async preview to ts and js templates because we have decent test coverage for compilation errors of those.

## Issue link

#331 
#276


alexfarrill (approved):
very nice!

PR #339, opened 2022-11-16
Replace poller with websockets

elliothursh:
## Describe your changes
Replaces the livereload polling implementation with sockets, fixes a bug where some file changes don't update the client.

## Issue link
N/A


packages/cli/src/commands/preview/server/__integration__/livereload.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      expect(WATCH_IGNORE.test(".next")).toBe(true);
+      expect(WATCH_IGNORE.test("node_modules")).toBe(true);
+    });
+
+    it("socket emits 'reload' when a template is touched", (done) => {
+      clientSocket.on("reload", (_arg: any) => {
+        done();
+      });
+
+      touchTemplate();
```
psugihara, line 37: nice

packages/cli/src/components/hooks/useLiveReload.tsx
```diff
@@ -1,42 +1,17 @@
...
+
+    socket.on("connect", () => {
+      console.info("Connected to live reload server");
+    });
+
+    socket.on("reload", () => {
+      console.info("Reloading...");
+      onShouldReload();
+    });
+  }, [onShouldReload]);
```
psugihara, line 16:
```suggestion
  return () => {
    socket.off();
    socket.disconnect();
  };
  }, [onShouldReload]);
```

I think you want to disconnect and unbind event listeners in a cleanup function. Otherwise this may leak memory.

packages/cli/src/components/hooks/useLiveReload.tsx
```diff
@@ -1,42 +1,17 @@
...
-    };
-  }, [vectorClock, onShouldReload]);
+    const socket = io();
+
+    socket.on("connect", () => {
+      console.info("Connected to live reload server");
+    });
+
+    socket.on("reload", () => {
+      console.info("Reloading...");
```
psugihara, line 13: rm chatty logging

packages/cli/src/components/hooks/useLiveReload.tsx
```diff
@@ -1,42 +1,17 @@
...
-    const interval = setInterval(checkForReload, LONG_POLLING_INTERVAL);
-    checkForReload();
-    return () => {
-      abortController.abort();
-      clearInterval(interval);
-    };
-  }, [vectorClock, onShouldReload]);
+    const socket = io();
+
+    socket.on("connect", () => {
```
psugihara, line 8: rm

psugihara (changes requested):
Feels nice and simple. Just a couple things to fix in the hook.

Will be interested in your review of using this tomorrow for the 0.9 email.

package.json
```diff
@@ -21,10 +21,10 @@
     "test": "yarn link:emails; yarn jest",
     "test:integration": "yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts $TEST_FLAGS\"",
     "test:integration:watch": "yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts --watch\"",
-    "ci:test": "yarn integration:servers:start test",
-    "ci:test:integration": "yarn integration:servers:start \"yarn link:emails; yarn jest -c jest.integration.config.ts\"",
     "test:integration:cypress": "yarn integration:servers:start \"cd packages/cli && yarn cypress run -C cypressIntegration.config.ts\"",
     "test:integration:cypress:headed": "yarn integration:servers:start \"cd packages/cli && yarn cypress run -C cypressIntegration.config.ts --headed\"",
+    "ci:test": "yarn integration:servers:start test",
```
alexfarrill, line 26: does this need to run servers now?  can it instead do the same thing as the "test" command?

alexfarrill (changes requested):
❌ it does not navigate between previews correctly
- https://www.loom.com/share/09d09bc28d6b4244a559752b9c507067
✅ it updates when editing a preview
✅ it updates when editing a template
✅ it updates when editing a component

packages/cli/src/components/hooks/useLiveReload.tsx
```diff
@@ -1,42 +1,21 @@
...
+      console.info("Connected to live reload server");
+    });
+
+    socket.on("reload", () => {
+      console.info("Reloading...");
+      onShouldReload();
+    });
```
psugihara, line 15:
```suggestion
    });
    checkForReload();
```

psugihara:
update here: this still errors on elliot's comp but not mine.

I think we found the race condition.... on a file change, mailling server re-bundles and tells the client to refresh. However the client refreshes by hitting a next /api route and that API route doesn't always have the latest bundle.

We're solving by adding an ID to the bundle and passing that ID to the API route, so that it can do a busy wait until the desired bundle ID is ready.

package.json
elliothursh: it still needs to run servers, there's still a few other tests that don't pass without them.

psugihara (approved):
👍 down to merge this duct-tape while we figure out #341

Issue #338, opened 2022-11-15
detect typescript in redwood project


alexfarrill:
mailing init currently detects typescript by looking for `./tsconfig.json`.  in a new redwood app, the tsconfig is in the `web` subdirectory.  make it so that we correctly detect typescript for redwood projects.

alexfarrill:
- [ ] add a test that reads mailing.config.json and fails if it is a typescript project but the "typescript" key is false (and vice versa)

PR #340, opened 2022-11-16
e2e test fixes/improvements

alexfarrill:
I peeled this off my other branch to make it easier to review, plus it corrects some bugs in main
In this branch:
- align "skipped apps" when running in localhost with what runs in github workflows currently (all frameworks but turbo are supported)
- restore "save-cache" and "use_cache" functionality
- introduce an `update-snapshot` cli command that updates the e2e jest snapshot tests

- next_js behavior: in main, the next apps have the wrong flags for typescript (default flipped in v13), so next_js likely does not install the javascript version of next. additionally, next no requires you to choose eslint or not, otherwise there is a prompt that causes the tests to run very slowly while it waits for the prompt to timeout
- redwood_ts behavior: in main, redwood_ts actually installs the jsx templates because tsconfig.json is not in the root directory so it's not detected as being a typescript by our current naive algo.  see https://github.com/sofn-xyz/mailing/issues/338 for the full fix.  this branch includes a temporary fix (override mailing.config.json with `npx mailing init --typescript=true`) the warning below that is now output when redwood_ts runs:
```
+        * * * * * * * * * * * * * * * * *
+        ⚠️ WARNING ⚠️ Expected mailing.config.json to have 'typescript' set to true but it was not!\n
+        In other words, mailing init did not correctly detect that this is a typescript project\n
+        Please implement https://github.com/sofn-xyz/mailing/issues/338 and then have this raise an error instead or move this check to a jest test\n
+        Until #338 is implemented, the e2e tests will override this by passing the --typescript flag to `npx mailing init` in typescript frameworks.
+        * * * * * * * * * * * * * * * * *
```

### DIFF
--- a/e2e/app.rb
+++ b/e2e/app.rb
@@ -10,6 +10,9 @@ require_relative 'app/remix_ts'
 require_relative 'app/remix_js'
 
 module App
+  # does this framework use typescript? if so, e2e tests will perform typescript specific checks
+  attr_reader :typescript
+
   CONFIGS = {
     standalone: App::Standalone,
     turbo: App::Turbo,
@@ -20,6 +23,8 @@ module App
     remix_ts: App::RemixTs,
     remix_js: App::RemixJs
   }.freeze
+
+  SKIPPED_APPS = %i[turbo].freeze
 
   def self.from_name(app_name)
     CONFIGS[app_name.to_sym]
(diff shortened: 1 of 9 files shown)


psugihara (approved):
love it. good finds

PR #347, opened 2022-11-18
Unsub page with single list has a button instead of a checkbox list

alexfarrill:
## Describe your changes

Unsub page with single list has a button instead of a checkbox list

Screenshot 2022-11-17 at 9 28 56 PM
Screenshot 2022-11-17 at 9 27 58 PM

This PR adds:
- A separate UI if you just have one list, with an Unsubscribe or Re-subscribe button (see screenshots)
- disable the save button while the form is saving to prevent race condition
- hide the "Saved!" message if it is showing when the user clicks Save
- Cypress tests for both UI's
- cy command: subscribeToDefaultList
- cy command: createList
- cy command: subscribe


alexfarrill:
why web-emails is failing agian?

psugihara (approved):

PR #359, opened 2022-11-22
357 previewsender should show a helpful message when your transport is not connected

alexfarrill:

### DIFF
--- a/packages/cli/src/components/PreviewSender.tsx
+++ b/packages/cli/src/components/PreviewSender.tsx
@@ -30,6 +30,7 @@ const PreviewSender: React.FC<PreviewSenderProps> = ({
 
       window.localStorage.setItem("previewSenderEmail", email);
       try {
+        setError(null);
         setSending(true);
         const payload: SendPreviewRequestBody = {
           to: email,
@@ -113,7 +114,7 @@ const PreviewSender: React.FC<PreviewSenderProps> = ({
           className="text-sm border-none p-[9px] rounded-r-sm cursor-pointer bg-[#fff] text-black hover:bg-white"
         />
       </form>
-      {error && !sending && (
+      {error && (
         <div className="error break-words max-w-[288px]">
           <span className="text-amber-300">⚠</span> {error}
         </div>
--- a/packages/cli/src/components/PreviewSender.tsx
+++ b/packages/cli/src/components/PreviewSender.tsx
@@ -12,19 +12,24 @@ const PreviewSender: React.FC<PreviewSenderProps> = ({
   previewClass,
 }) => {
   const [email, setEmail] = useState<string | null>(null);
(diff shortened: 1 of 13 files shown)


packages/cli/src/components/PreviewSender.tsx
```diff
@@ -12,19 +12,24 @@ const PreviewSender: React.FC<PreviewSenderProps> = ({
...
       const lastSent = window.localStorage.getItem("previewSenderLastSentAt");
       if (lastSent) {
         setLastSentAt(new Date(lastSent));
       }
+
+      // ensure this only happens once
+      setUsedEmailFromLocalStorage(true);
     }
-  }, [email]);
+  }, [email, usedEmailFromLocalStorage]);
```
alexfarrill, line 32: There's a way to do this with less code by passing [] (https://reactjs.org/docs/hooks-effect.html), but I would have to disable the tslint for this line:  "React Hook useEffect has missing dependencies: 'email' and 'usedEmailFromLocalStorage'. Either include them or remove the dependency array."

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.

wdyt?

packages/cli/src/components/PreviewSender.tsx
```diff
@@ -98,10 +113,20 @@ const PreviewSender: React.FC<PreviewSenderProps> = ({
...
+        <input
+          type="submit"
+          value="Send"
+          disabled={!email?.length}
+          className="text-sm border-none p-[9px] rounded-r-sm cursor-pointer bg-[#fff] text-black hover:bg-white"
         />
-        <input type="submit" value="Send" disabled={!email?.length} />
       </form>
-      {error && <div className="error">⚠ {error}</div>}
+      {error && !sending && (
```
alexfarrill, line 125: clear the error message when you send again

packages/cli/src/components/PreviewSender.tsx
psugihara: would this work?

```
  useEffect(() => {
       setEmail(window.localStorage.getItem("previewSenderEmail"));
       const lastSent = window.localStorage.getItem("previewSenderLastSentAt");
       if (lastSent) {
         setLastSentAt(new Date(lastSent));
       }
   }, []);
```

psugihara (approved):

packages/cli/src/components/PreviewSender.tsx
alexfarrill: yes, but ts lint does not allow

alexfarrill (commented):
Nice

alexfarrill:
JSONified error message from the server:

Screenshot 2022-11-22 at 4 22 49 PM

PR #365, opened 2022-11-23
Await create default list

alexfarrill:
## Describe your changes

- bugfix missing await
- add lint rule for missing await
- (unrelated) also assert organization is created after user create

Screenshot 2022-11-23 at 1 30 28 PM

Screenshot 2022-11-23 at 1 29 44 PM


psugihara (approved):
lol!

PR #370, opened 2022-11-24
Lint no floating promises 2

alexfarrill:


packages/cli/src/commands/init.ts
```diff
@@ -86,7 +86,7 @@ export const handler = buildHandler(
           try {
             const url = `${getMailingAPIBaseURL()}/api/newsletterSubscribers`;
 
-            fetch(url, {
+            await fetch(url, {
```
psugihara, line 89: here we're purposefully not awaiting bc we don't care about the response and don't want to make the user wait. Is that what `void` is for?

packages/cli/src/pages/api/lists/index.ts
```diff
@@ -71,10 +71,10 @@ const ApiLists = withSessionAPIRoute(async function (
 
   switch (req.method) {
     case "GET":
-      handleGetLists(user, req, res);
+      await handleGetLists(user, req, res);
       break;
     case "POST":
-      handleCreateList(user, req, res);
+      await handleCreateList(user, req, res);
```
psugihara, line 77: nice, this probably causes occasional errors

psugihara (commented):
Looks like this caught good stuff. I'm going to try it locally.

packages/cli/src/commands/init.ts
alexfarrill: Yeah, i will change it to void.  If we can't get the linting to work, we could just pull out the changes that it turned up and do that for now

package.json
```diff
@@ -9,7 +9,9 @@
     "packages/*"
   ],
   "scripts": {
-    "lint": "eslint .",
+    "lint": "yarn lint:unlink; eslint .; yarn lint:link",
+    "lint:unlink": "rm packages/cli/src/emails",
```
elliothursh, line 13: why are these changes necessary?

alexfarrill: The symlink causes the linter to misbehave.  Basically, the linter complains that each file in packages/cli/src/emails is not covered by a tsconfig file, but the path it lists is the resolved path of each file in generator_templates/ts/emails/.  There is no way to indicate to the linter to ignore the symlink because it first attempts to resolve the symlink, so you can either do this or you have to tell the linter to ignore the files that are in generator_templates/ts/emails/.  You can try removing this and running the linter to see the error messages

packages/cli/src/components/PosthogScript.tsx
```diff
@@ -1,4 +1,4 @@
-import { POSTHOG_API_KEY } from "../util/postHog/client";
+import { POSTHOG_API_KEY } from "../util/postHog/posthogApiKey";
```
psugihara, line 1: really would have thought tree shaking would work here 🤷

psugihara (approved):
lgtm!

packages/cli/src/components/PosthogScript.tsx
alexfarrill: Yeah, it is incredible (-y bad) that it doesn't!

PR #377, opened 2022-11-28
render.ts - api key auth via process.env

alexfarrill:
from example PR provided

### DIFF
--- a/packages/cli/src/pages/api/render.ts
+++ b/packages/cli/src/pages/api/render.ts
@@ -16,9 +16,7 @@ export default async function handler(
 ) {
   const { templateName, props } = "GET" === req.method ? req.query : req.body;
 
-  // if process.env.REQUIRE_API_KEY is set then require a valid api key to be passed in
-  if (process.env.REQUIRE_API_KEY === "true") {
-    if (!(await validateApiKey(req, res))) {
+  if (process.env.REQUIRE_API_KEY === "true" && !(await validateApiKey(req, res))) {
       return res.status(401).json({ error: "API key is not valid" });
     }
   }
--- a/packages/cli/src/pages/api/render.ts
+++ b/packages/cli/src/pages/api/render.ts
@@ -2,7 +2,7 @@ import type { NextApiRequest, NextApiResponse } from "next";
 import { MjmlError } from "mjml-react";
 
 import renderTemplate from "../../util/renderTemplate";
-import { validateApiKey } from "src/util/validateApiKey";
+import { apiKeyFromReq } from "src/util/validateApiKey";
 
 type Data = {
   error?: string; // api error messages
@@ -10,13 +10,20 @@ type Data = {
   mjmlErrors?: MjmlError[];
 };
 
-export default async function handler(
+export default function handler(
   req: NextApiRequest,
   res: NextApiResponse<Data>
 ) {
   const { templateName, props } = "GET" === req.method ? req.query : req.body;
 
-  if (!(await validateApiKey(req, res))) return;
+  // if process.env.MAILING_API_KEY is set then require an api key that matches to be passed in
+  if (process.env.MAILING_API_KEY) {
+    const apiKey = apiKeyFromReq(req);
+
+    if (apiKey !== process.env.MAILING_API_KEY) {
+      return res.status(401).json({ error: "API key is not valid" });
+    }
+  }
 
   // validate template name
   if (typeof templateName !== "string") {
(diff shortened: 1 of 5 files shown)


packages/cli/src/pages/api/render.ts
```diff
@@ -2,19 +2,22 @@ import type { NextApiRequest, NextApiResponse } from "next";
...
 };
 
-export default function handler(
+export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse<Data>
 ) {
   const { templateName, props } = "GET" === req.method ? req.query : req.body;
 
+  if (!(await validateApiKey(req, res))) return;
```
psugihara, line 19: can we put this check behind a config flag? I don't really think we should validate for render by default and don't want to break any existing users.

Maybe 

```
  if (process.env.MAILING_REQUIRE_API_KEY && !(await validateApiKey(req, res))) return;
```

psugihara (changes requested):
looks good, one thing.

Where should we add this to the docs?

packages/cli/src/pages/api/render.ts
alexfarrill: I was thinking the same thing, good suggestion

alexfarrill:
@psugihara have another look, i think it makes sense for render to adopt this behavior if MAILING_API_KEY is set, otherwise not.  and for the validation logic to live separately from the db api key validation (right now they exist as separate features).  wdyt?

psugihara:
Tests look good but I think it's kind of confusing that this api key would work for api/render but not e.g. api/sendMail. With this setup there are 2 kinds of API keys that do 2 different things. I think if I'm setting the api key, it should work as the api key in all contexts. That's simpler conceptually for me.

alexfarrill:
Makes sense. I can flip it to be one function. Do you think the REQUIRE_API_KEY flag should only affect this one type of auth then or apply to all routes and all methods of auth?

psugihara:
Cool. I think we just want REQUIRE_API_KEY to turn on the feature for /api/render (maybe there will be others in the future). Maybe there is a better name for that key.

alexfarrill:
@psugihara updated and added doc

packages/web/pages/docs/rest-api.mdx
```diff
@@ -28,3 +34,10 @@ With the REST API, you can use mailing for email templating even if most of your
 | POST   | /api/lists/[listId]/members            | add a member to a list        |
 | GET    | /api/lists/[listId]/members/[memberId] | get a list member's status    |
 | PATCH  | /api/lists/[listId]/members/[memberId] | update a list member's status |
+
+## API keys
+
+For endpoints that require an API key, the following methods are supported:
+
+- Include the API key in the request headers: `X-API-KEY=...`
+- Include the API key as a query string parameter: `?apiKey=...`
```
psugihara, line 43:
```suggestion
## API keys

Mailing uses API keys to protect endpoints that consume limited resources (e.g. sending email with your transport).

To get an API key, either [set up a database](https://www.mailing.run/docs/deploy#adding-a-database) and log in (recommended) or set the `MAILING_API_KEY` environment variable in your deployment environment.

You then pass your API key in each request with one of these methods:

- Include the API key in the request headers: `X-API-KEY=...`
- Include the API key as a query string parameter: `?apiKey=...`


```

packages/cli/src/pages/api/render.ts
```diff
@@ -2,19 +2,27 @@ import type { NextApiRequest, NextApiResponse } from "next";
...
-export default function handler(
+export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse<Data>
 ) {
   const { templateName, props } = "GET" === req.method ? req.query : req.body;
 
+  // if process.env.REQUIRE_API_KEY is set then require a valid api key to be passed in
+  if (process.env.REQUIRE_API_KEY === "true") {
+    if (!(await validateApiKey(req, res))) {
```
psugihara, line 21:
```suggestion
  if (process.env.REQUIRE_API_KEY === "true" && !(await validateApiKey(req, res))) {
```

psugihara (approved):
looks good. Added to the doc a bit to tell you how to get an API key.

packages/web/pages/docs/rest-api.mdx
alexfarrill: Is it clear enough that you need to make up your own API key if you set MAILING_API_KEY?  We might get questions about where to get a key from

psugihara: 👍  edited again for clarity. feel free to edit

alexfarrill: Nice 👍

PR #379, opened 2022-11-29
Sync Templates

elliothursh:
## Describe your changes
This generates javascript template files from typescript template files as a pre-commit hook if any of the typescript template files change. This will allow us to more easily make changes to template files without having to remember to make changes in both ts and js files. 

~~The only caveat is that the Javascript template files do not retain empty lines from the typescript file.~~ Newlines are kept after imports thanks to an eslint rule.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,4 @@ node_modules
 **/dist/**
 **/generated/**
 **/runs/**
+.mailing
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -69,6 +69,12 @@
         "@next/next/no-html-link-for-pages": "error",
         "@next/next/no-img-element": "error"
       }
+    },
+    {
+      "files": ["**/moduleManifest.js"],
+      "rules": {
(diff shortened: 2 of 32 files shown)


psugihara (commented):
very sick! Impressed with how clean the js comes out. We hadn't done this because we thought it would be mangled, but you found good tsc settings 🙌 

Is there a way to keep the newlines? That's the only bit that looks odd... never have I ever seen someone not do a newline after imports.

It looks like [prettier won't do it](https://github.com/prettier/prettier/issues/8321) but it does look like there's an [eslint rule we could use](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/newline-after-import.md).

elliothursh:
Oh nice find with the eslint rule, works well!

alexfarrill (approved):
It would be cool if the jsx could have a comment that they're generated not to edit them, but that shouldn't carry through to the build.  But not sure it's worth the extra effort right now, I think it's good as is

psugihara (approved):
beautiful

PR #385, opened 2022-11-29
use 422 response codes

alexfarrill:
In this update:
- sendMail returns 422 response codes instead of 403
- change the response code from 404 to 422 if you pass an invalid template name
- the http responses for an invalid api key to use json instead of plaintext
- adds tests for the above


elliothursh (approved):
sweet!

PR #388, opened 2022-11-30
Improvements to sendMail

alexfarrill:
In this PR:
- sendMail throws error message that include a "status" attribute that can be used as an status code in the context of an http request.
- /api/sendMail uses component rendering instead of pre-rendering to html.  This allows us to delete some redundant logic in the controller and fixes a bug where it would not use the subject function of the template being used
- tests

### DIFF
--- a/packages/cli/src/pages/api/__test__/sendMail.test.ts
+++ b/packages/cli/src/pages/api/__test__/sendMail.test.ts
@@ -29,66 +29,116 @@ function mockRequestResponse(method: string) {
 }
 
 describe("sendMail with a valid api key", () => {
-  it("should 422 if no to, cc, or bcc is provided", async () => {
-    const { req, res } = mockRequestResponse("POST");
-    req.body = {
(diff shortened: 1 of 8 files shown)


packages/cli/src/pages/api/__test__/sendMail.test.ts
```diff
@@ -28,67 +20,26 @@ function mockRequestResponse(method: string) {
...
 }
 
-describe("sendMail with a valid api key", () => {
-  it("should 422 if no to, cc, or bcc is provided", async () => {
-    const { req, res } = mockRequestResponse("POST");
-    req.body = {
-      templateName: "AccountCreated",
-      props: { name: "Alex" },
-    };
+describe("sendMail with a valid api key and real sendMail -- be careful as this could send email if tranpsort is configured", () => {
```
psugihara, line 23: typo - "transport"

psugihara (approved):
nice tests

PR #393, opened 2022-12-01
fix: prevent intercept error when `from` is object

liamross:
## Describe your changes

This prevents the intercept page breaking when `from` is an object instead of a string.

When you call `sendMail` like this:
```ts
sendMail({
  from: { name: "Some name", address: "some.name@example.com" },
});
```

It breaks when intercepted:

```console
Uncaught Error: Objects are not valid as a React child (found: object with keys {name, address}). If you meant to render a collection of children, use an array instead.
```

NOTE: I wasn't sure how to actually test this locally, so hopefully my changes are good. Let me know if there's an easy way to send emails for interception.

## Issue link

N/A

### DIFF
--- a/.changeset/sweet-berries-know.md
+++ b/.changeset/sweet-berries-know.md
@@ -1,0 +1,6 @@
+---
+"mailing": patch
+"mailing-core": patch
+---
+
+fix: prevent intercept error when from is object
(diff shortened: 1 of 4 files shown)


psugihara:
Tested manually ✅. Great catch, thanks so much!

alexfarrill:
Thanks @liamross !

Issue #188, opened 2022-09-12
Port `3883` is hard-coded during interception


iDuuck:
Just have a look at the following snippet: https://github.com/sofn-xyz/mailing/blob/b0f4c800b19a1f757ae962d1f840d4d645fca9bb/packages/core/src/index.ts#L117

Having the server running on a different port results in interceptions not working properly.

alexfarrill:
Thanks @iDuuck for opening an issue about this -- we've known about this problem and you're right, it's tricky to support a custom port for the reasons you mention.  One option could be to allow you to change the port in the configuration file, but not from a CLI option.  Would that work for you?

If you have a moment to tell us a bit more about how you're using mailing, we'd love to hear more about it in this thread: https://github.com/sofn-xyz/mailing/discussions/121

PR #394, opened 2022-12-01
fix: allow custom preview server url

liamross:
## Describe your changes

At the moment the intercept URL is hard coded, despite allowing the preview server to live at a different port via the --port flag

NOTE: My understanding is that adding this variable is necessary since I don't see a way to propagate the --port flag for the server over to the `mailing-core` package (I could be wrong, if so feel free to overwrite / let me know so I can fix).

ALSO NOTE: I did not test this again, apologies I just thought this would be higher value than opening up an issue and didn't want to dig too deep.

## Issue link

N/A

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -253,7 +253,7 @@ export function buildSendMail<T>(options: BuildSendMailOptions<T>) {
   };
 }
 
-async function openPreview(
+export async function openPreview(
   mail: ComponentMail,
   mailOptions: SendMailOptions,
   previewServerUrl?: string
@@ -278,6 +278,7 @@ async function openPreview(
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify(mailOptions),
     });
+
     if (response.status === 201) {
       const { id } = (await response.json()) as { id: string };
       await open(`${PREVIEW_SERVER_URL}/${id}`);
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -33,6 +33,7 @@ export type ComponentMail = SendMailOptions & {
   component?: JSX.Element;
   dangerouslyForceDeliver?: boolean;
   forcePreview?: boolean;
+  previewServerUrl?: string;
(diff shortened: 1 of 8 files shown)


packages/core/src/index.ts
```diff
@@ -164,7 +166,7 @@ export function buildSendMail<T>(options: BuildSendMailOptions<T>) {
       log("💌 opening sendMail preview", debugMail);
       // create an intercept on the preview server
       // then open it in the browser
-      const PREVIEW_SERVER_URL = "http://localhost:3883/intercepts";
+      const PREVIEW_SERVER_URL = `${previewServerUrl?.trim().replace(/\/+$/, "") || "http://localhost:3883"}/intercepts`;
```
liamross, line 169: This regex replaces trailing slashes so this doesn't break if they are included.

Demo: regexr.com/73m23

packages/web/pages/docs/sending-email.mdx
```diff
@@ -62,6 +62,7 @@ Sends an email using the configured transport. It extends the `sendMail` functio
 - `component` – The React component to render as the email body. If you pass in a string for `html`, this will be ignored
 - `dangerouslyForceDeliver` – If true, the email will be sent even if the `NODE_ENV` is not `production`. Use with caution.
 - `forcePreview` – If true, the email will be intercepted by the preview server
+- `previewServerUrl` – If you have changed the port of the mailing preview server, you can set it here (default: `http://localhost:3883`)
```
liamross, line 65: Not sure about this copy, just wanted to put something in here.

psugihara:
Thanks, I like this as a band-aid for https://github.com/sofn-xyz/mailing/issues/188

Will try to take a look soon to add tests.

- [ ] add automated test that previewServer is used

alexfarrill:
I did a small refactor and added some tests, have a look :)

psugihara (approved):
cool!

alexfarrill:
Thanks for the PR @liamross , we just merged it in!

PR #401, opened 2022-12-05
include mailing_version in posthog analytics

alexfarrill:


alexfarrill:
Verified:

- [x] (mailing cli) released 0.9.5-next.2, included in a demo project and the events include mailing_version:
image
- [ ] (mailing frontend)
- [ ] (mailing core)

alexfarrill:
Verified:

- [x] (mailing cli) released 0.9.5-next.2, included in a demo project and the events include mailing_version:
image

- [x] (mailing frontend - dev frontend does not include posthog script)

- [ ] (mailing frontend - production vercel deploy includes posthog script)

- [ ] (mailing core)

alexfarrill:
Everything is working here now, but this was a pain in the ass.  Since scripts/release is getting kind of big now, some notes for a future refactor that could improve it:

- The constants that get replaced during the release process could all be one file per package, e.g. `src/const/dynamicConstants.ts` so we don't have to search and replace on every file and so you remember not to mess with the syntax of that file too much
- There should be some validation that each search found something to replace and that it was replaced back to the way it was before.  One idea for a foolproof restore is you could use git: require a clean git status when release is run, then do your search and replace, afterwards git reset --hard (or git stash the changes, then you could inspect what the script did too)
- If the script gets much more complicated, we could rewrite it in ruby to make it easier to perform some of the above checks and to manage it

psugihara (approved):
lgtm 👍

PR #410, opened 2022-12-07
fix #407 preview server in mailing 094 crashes because of posthog shutdown

### LINKED ISSUE #407: preview server in mailing 0.9.4 crashes with [CanceledError: canceled] { code: 'ERR_CANCELED' }
- 0.9.3 does not crash
- the browser does not have to be open for the server to crash

commit that adds the linting in 0.9.4: `4fe6211ac72de5ed2e9f1e7c78913a9fba76fb4e`

suspects:
- packages/cli/src/commands/init.ts added an await for previewHandler
- packages/cli/src/util/buildHandler.ts added an await for shutdownAnalytics()

alexfarrill:
- upgrade posthog-node to 2.2.3
- wrap posthog client shutdown in a try/catch and test

### DIFF
--- a/.changeset/large-monkeys-count.md
+++ b/.changeset/large-monkeys-count.md
@@ -1,0 +1,7 @@
+---
+"mailing": patch
+"mailing-core": patch
+"web": patch
+---
+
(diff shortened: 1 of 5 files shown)


elliothursh (approved):
cool 👍

psugihara (approved):
great detective work!

PR #418, opened 2022-12-09
sendMail uses anonymousId

alexfarrill:
simple made easy


psugihara (approved):
a classic! https://www.infoq.com/presentations/Simple-Made-Easy/

Issue #427, opened 2023-01-22
Body exceeded 1mb limit


rafyzg:
**Is your feature request related to a problem? Please describe.**
I am trying to send a mail with attachment of file size larger than 1mb
I get the following message:
[Body exceeded 1mb limit](https://stackoverflow.com/questions/68574254/body-exceeded-1mb-limit-error-in-next-js-api-route)

As you see in the link attached this is an issue with the default size limit in next
I am not sure if currently I have an option to increase the size limit.

**Describe the solution you'd like**
Ability to increase the body size limit.
Two options:
1. increase the size limit for sendMail route  (I could push a fix if necessary) 
2. next.config.js file where you could extend default settings of next..?

**Describe alternatives you've considered**
Currently I don't have other ones, I am open for ideas.

psugihara:
I'm up for increasing the size limit and would be happy to review a PR. Thank you for making the issue! That said, are you trying to use sendMail via the REST API? As far as I know, that does not support attachments.

rafyzg:
It does support. 
https://www.mailing.run/docs/sending-email#common-componentmail-options

I will submit a PR in couple of days. 
Thanks.

PR #428, opened 2023-01-22
Increase sizeLimit in sendMail route

rafyzg:
## Describe your changes
Added config object to sendMail route to increase default sizeLimit

## Issue link
https://github.com/sofn-xyz/mailing/issues/427

- [ V] I have performed a self-review of my code


rafyzg:
I see that tests are failing. 
Is it there something I should do from my side?

psugihara:
No, I think it's not from this change. Going to merge this and I'll get those fixed before releasing.

Thanks much, @rafyzg!

PR #444, opened 2023-02-22
example .env

alexfarrill:


alexfarrill:
@elliothursh @psugihara any other env variables you use in mailing?

psugihara (approved):
looks good to me. I don't think this needs a changeset though since it doesn't require a version bump.

Issue #429, opened 2023-01-23
Feature request: password protected preview site


maxsalven:
I'd love to be able to password protect the entire preview site when I deploy it to production. I see that protections are available for the API routes but anyone who knows my Vercel URL can see all preview emails. I'd like to limit this to approved collaborators through a password.

alexfarrill:
Hey there Max, this would be a welcome feature and we would be happy to review a PR. Thanks for making the issue!

jpalumickas:
Hey, also would be nice to disable previews with ENV variable (for example `DISABLE_PREVIEWS=true`) in production. Maybe this would be easer to implement?

alexfarrill:
@jpalumickas Thanks for the idea! cc: @maxsalven 

I've released a beta version `0.9.10-next.0` that implements this.  If you set `DISABLE_PREVIEWS` to something truthy, previews will be disabled.

Would you try it out and let me know if this solves your problem?

Thanks

jpalumickas:
@alexfarrill it's working! maybe `DISABLE_PREVIEWS` check should be added to `pages/index.ts` to not do redirection to previews?

alexfarrill:
@jpalumickas  @maxsalven This is released in 0.9.11.  Thank you again for using Mailing!

PR #445, opened 2023-02-24
DISABLE_PREVIEWS env variable

alexfarrill:
idea from #429


psugihara (approved):
good idea.

We should put this in the docs somewhere too.

PR #459, opened 2023-03-08
Framework test improvements and bump node

alexfarrill:
- bump node 16.18.0 -> 16.19.0 because latest create-redwood-app requires 16.19.0 at minimum

Framework test bumps:
- update create-next-app defaults so that the user is not prompted for more info
- update create-redwood-app defaults so that the user is not prompted for more info
- split standalone into js and ts (will help to debug @faire/mjml-react branch)


psugihara (approved):
awesome

PR #466, opened 2023-04-20
Pass minifier callback to mjml-react render method

Jaraujo6:
## Describe your changes
Currently, the `sendMail` api turns off minification by default and does not provide a mechanism by which users can supply their own minification. This pull request seeks to change that by adding an optional callback to the `buildSendMail` option arg, that if present, will be called using html created from the `react-mjml` render method, and returned from the `mailing` render method. 

This approach decouples the minifier dependency from the larger mailing repository which means that users can use any library or implementation they please, as long as the callback accepts a string as an argument and returns a string, reducing the maintenance burden on the mailing support team.

## Issue link
N/A

- [ x ] I have performed a self-review of my code

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -27,7 +27,7 @@ export type BuildSendMailOptions<T> = {
   transport: Transporter<T>;
   defaultFrom: string;
   configPath: string;
-  minifyCb?: (html: string) => string;
+  processHtml?: (html: string) => string;
 };
 
 export type ComponentMail = SendMailOptions & {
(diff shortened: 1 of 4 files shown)


psugihara (commented):
looks good, just tweaked the naming to be one step more general

packages/core/src/index.ts
```diff
@@ -27,6 +27,7 @@ export type BuildSendMailOptions<T> = {
   transport: Transporter<T>;
   defaultFrom: string;
   configPath: string;
+  processHtml?: (html: string) => string;
```
Jaraujo6, line 30: @psugihara should this follow the same `options` api you did in the render or do you think this is fine?

btw changing the name was the right call. i had similar thoughts so i'm glad you thought to do this

psugihara: I think this is fine because it's already in an options object. My reason for adding that was just in case we want to add more options in the future without breaking users who might be relying on the current type signature for `render`.

psugihara (approved):
thanks much @Jaraujo6 !

PR #468, opened 2023-05-15
the api should 500 not 401 if there is a db error (aside from not fin…

alexfarrill:


alexfarrill:
@psugihara I had to bump prisma in order to use the error codes, but figured that was safe enough

alexfarrill:
this branch is green aside from redwood_ts e2e that is failing in main because create redwood app does not work correctly.  i am disabling that test

psugihara (approved):
good fix!

Issue #479, opened 2023-09-08
/api/sendMail only sends with `defaultFrom`


maxaggedon:
**Describe the bug**
`from` option is correctly shown in intercept but not used

**To Reproduce**

1. Use /api/sendMail endpoint with a `from`
2. Force send if in preview mode
3. Actual email has the `defaultFrom` value as `From` header

**Expected behavior**
If specified, `from` field should be used

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Chrome
- Version: 1.0.1

PR #480, opened 2023-09-11
include `from` option when forcing in intercept preview

maxaggedon:
## Describe your changes
in preview mode, when a call to `/api/sendMail` is intercepted and the user clicks 'force send', the `from` option is now passed along

## Issue link
#479 


psugihara (approved):
sorry for delay! this looks great. thank you much for the contribution.

Issue #464, opened 2023-04-05
Sidebar sometimes disappears and gets stuck in faulty state


KATT:
**Describe the bug**

**To Reproduce**

Repro at https://github.com/KATT/mailing-run-issue

1. ```sh
    git clone git@github.com:KATT/mailing-run-issue.git
    cd mailing-run-issue
    yarn
    npx mailing
   ```
1. Click Welcome -> Preview email
2. Reload page

**Expected behavior**

Should look the same.

**Actual behavior**

- Sidebar disappears
- Gets stuck in faulty state even if you clear the URL

Workaround: clear the URL except for the `/preview` part

**Video**



https://user-images.githubusercontent.com/459267/230070695-4a94812d-43da-43bb-bada-e2d7a471d5dc.mp4


**Additional context:**

Might be because of the space here: https://github.com/KATT/mailing-run-issue/blob/aa8fb7744a0c8e13a28198355906c9aae636dbcb/emails/previews/Welcome.tsx#L4

marcincodes:
@psugihara This issue can be closed. Fix was merged here: https://github.com/sofn-xyz/mailing/pull/482

PR #482, opened 2023-09-30
Decode URI component when reading URL path

marcincodes:
## Describe your changes
Decode URI components when reading path to later find proper cursor fro tree view

## Issue link
https://github.com/sofn-xyz/mailing/issues/464


psugihara (approved):
looks good to me, seems to work. Thank you!

PR #488, opened 2023-11-26
fix: bump to next 14 in peer deps

KATT:
## Describe your changes

Bump to next 14 peer deps


packages/cli/package.json
```diff
@@ -42,7 +42,7 @@
     "node-mocks-http": "^1.11.0"
   },
   "peerDependencies": {
-    "next": "^12.0.0 || ^13.0.0",
+    "next": "^12.0.0 || ^13.0.0 || ^14.0.0",
```
KATT, line 45: I kinda think this could be

```suggestion
    "next": ">= 12.0.0",
```

They're pretty good at backwards compat and likely won't drop the `/pages`-router for a long time

